PR #38, opened 2015-11-30
Update README.md

RocketD0g:
Added 5 criteria for data sources that can be aggregated onto platform. 

@jflasher (and anyone else who would like), would appreciate you reviewing this first.


RocketD0g:
@jflasher (and anyone else who would like), please review.

PR #73, opened 2016-02-12
Replace MongoDB with PostgreSQL in the Readme file #72

dolugen:


dolugen:
I probably didn't need to create #72

jflasher:
Awesome, thanks. I made these changes pretty quickly, so I am sure all this documentation could be better.

Issue #76, opened 2016-02-15
Kimono is shutting down in 2 weeks, move off of Kimono


jflasher:
https://www.kimonolabs.com/

This is actually a good thing since it's been breaking on a regular basis anyways.

dolugen:
So it was a web scraping API? How are we going to replace this?

olafveerman:
I plan to work on this after the Iranian scraper. That one is almost done,
just need to do some research on averaging period and units of measurement.

Haven't looked into the sources that rely on Kimono, but most of the other
sources are now scraped using Cheerio.

RocketD0g:
For the Kimono stuff: We will need to re-build the scrapers for: New Delhi
(not the US Embassy site), US Embassy and Consulate reporting sites in
China, and London. The US Embassy and Consulate sites will be relatively
easy and can follow the same form as the scrapers already existing for
India, Vietnam, Indonesia, and Mongolia (see:
https://github.com/openaq/openaq-fetch/blob/master/adapters/stateair.js).
The other two will be more challenging.

In general, while annoying to have to re-build the scrapers, the Kimono
service was not dependable and often didn't work on weekends or for several
days in a row, so probably for the best.

On Sun, Feb 21, 2016 at 8:37 PM, Dolugen Buuralda <notifications@github.com>
wrote:

> So it was a web scraping API? How are we going to replace this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openaq/openaq-fetch/issues/76#issuecomment-186964587>.
>



-- 

507.246.2097

dolugen:
> The US Embassy and Consulate sites will be relatively easy and can follow the same form as the scrapers already existing for India, Vietnam, Indonesia, and Mongolia

I took the easier one and updated the `beijing.js` adapter and the related sources.

dolugen:
@RocketD0g Thanks for summarizing it like that.

Adapters that need to be updated:
- [x] Beijing
- [ ] India
- [ ] London

jflasher:
Beijing update looks good, merging in. Thanks @dolugen. Going to close this in favor of having #82 and #81 open.

PR #77, opened 2016-02-22
Update adapter and source for China

dolugen:
#76 

My attempt to update the adapter and sources for StateAir.net's Mission China. It's based on `stateair.js`.


dolugen:
Gotta recreate the PR now.

Issue #82, opened 2016-02-24
Move several Indian sources off Kimono


jflasher:
If someone grabs this, please assign it to yourself.

dolugen:
I'll take this.

PR #85, opened 2016-02-24
Update adapter and source for India

dolugen:
#82 

@jflasher Scraping is great and all, but it's susceptible to minor changes in the source resource. How are we handling that?

### DIFF
--- a/adapters/india.js
+++ b/adapters/india.js
@@ -1,14 +1,13 @@
 /**
- * This code is responsible for implementing all methods related to fetching
- * and returning data for the Indian Kimono data sources.
- *
- * @todo This should most likely be changed from Kimono in the future.
+ * This code is responsible for implementing all methods related to
+ * fetching and returning data for the Indian data sources.
  */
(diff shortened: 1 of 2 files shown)


jflasher:
Thanks @dolugen, I'll check this out.

For the scraping stuff, you're totally right that it is always breakable. Luckily, it doesn't look like these sites are changing so much (which is either good or bad!). I also try and generally make the scrapers 'break' if things change too much. Better to have no data than wrong data. We also are making sure all the data that gets saved is in the right format. But yeah, if a unit all of the sudden changes on a site, we may never know. Also, if we didn't have to make all these scrapers, what would we do in our spare time? :sob:

dolugen:
Would it be useful to save up the original source resources to avoid losing data and post-process them with updated adapters? It could be kept for a short time and cleaned automatically. Just an idea.

jflasher:
Pulled in with https://github.com/openaq/openaq-fetch/pull/92

PR #96, opened 2016-02-29
Hochiminhtest

RocketD0g:
Added Ho Chi Minh City. @jflasher, check it pls.

### DIFF
--- a/adapters/stateair.js
+++ b/adapters/stateair.js
@@ -42,6 +42,7 @@ var formatData = function (data) {
         case 'New Delhi':
           return 'Asia/Kolkata';
         case 'Hanoi':
+        case 'Ho Chi Minh City':
           return 'Asia/Ho_Chi_Minh';
         case 'Ulaanbaatar':
           return 'Asia/Ulaanbaatar';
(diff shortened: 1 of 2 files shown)


adapters/stateair.js
```diff
@@ -42,6 +42,7 @@ var formatData = function (data) {
         case 'New Delhi':
           return 'Asia/Kolkata';
         case 'Hanoi':
+        case 'Ho Chi Minh City':
```
jflasher: switch case fallthrough :boom:

sources/vn.json
```diff
@@ -10,5 +10,18 @@
...
         ]
+    },
+    {
+        "url": "http://stateair.net/dos/RSS/HoChiMinhCity/HoChiMinhCity-PM2.5.xml",
+        "adapter": "stateair",
+        "name": "Ho Chi Minh City",
+        "city": "Ho Chi Minh City",
+        "country": "VN",
+        "description": "",
+        "resolution": "1 hr",
```
jflasher: You can get rid of `resolution`, we don't use it anymore.

PR #107, opened 2016-03-06
Don't use Dutch locale for date parsing

### LINKED ISSUE #106: No data for NL since a week ago
https://api.openaq.org/v1/latest?country=NL No data since Feb 29th. Source is still updated regularly http://www.lml.rivm.nl/sos/?C=M;O=D

Local run of `node index.js --source Netherlands` says: "New measurements inserted for Netherlands: 0"

@jflasher Any idea what's going on?

jflasher:
fixes #106

### DIFF
--- a/adapters/netherlands.js
+++ b/adapters/netherlands.js
@@ -70,7 +70,7 @@ var listApachetree = function (data, url) {
   var $ = cheerio.load(data);
 
   var parseDate = function (ds) {
-    var date = moment.tz(ds, 'DD-MMM-YYYY HH:mm', 'nl', 'Europe/Amsterdam');
+    var date = moment.tz(ds, 'DD-MMM-YYYY HH:mm', 'Europe/Amsterdam');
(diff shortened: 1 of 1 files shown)


jflasher:
This should fix the NL no data issue. We were using a NL locale, but it looks like the file list at http://www.lml.rivm.nl/sos/ does not use Dutch date names so we were missing anything in March. This also means we probably didn't get anything new in October. Looks like we're also not using any NL locale anywhere in the adapter now, which I think is fine?

cc/ @olafveerman @dolugen

dolugen:
Huh, I was wondering what that `nl` parameter was for.

olafveerman:
This makes me wonder if they changed their server settings, I'm pretty sure this was working well before

dolugen:
@olafveerman Can you point me to the reference where you use locale in momentjs? I can't seem to find such example in the package docs.

PR #108, opened 2016-03-06
Adding AirNow data

jflasher:
- Removing previous Texas adapter as it was only for Houston data, this encompasses that data as well


dolugen:
So we won't be able to test this adapter locally?

jflasher:
You can create a free account at https://docs.airnowapi.org/

PR #110, opened 2016-03-07
Add migration to fix NL local dates

### LINKED ISSUE #91: Correct inaccurate historic NL measurements
They're off by an hour for dates before today. Will need to write a script to be run once to fix this. Will need to update JSON in the `data` field.

dolugen:
Fixes #91.


jflasher:
:+1:

PR #167, opened 2016-05-13
Add missing _id in select query

dolugen:
Otherwise the rollback function wouldn't work as intended.

It's already migrated, but...


jflasher:
Good catch!

PR #173, opened 2016-06-02
Spartan data

jflasher:
Looks like there are about 110,000 measurements to get pulled in. We should do some spot checking on the data to make sure it looks good with the files, bringing in PM25 and BC data.

Data to be inserted into system is below (to compare to source data).
[output.txt.zip](https://github.com/openaq/openaq-fetch/files/296528/output.txt.zip)

### DIFF
--- a/adapters/spartan.js
+++ b/adapters/spartan.js
@@ -1,0 +1,323 @@
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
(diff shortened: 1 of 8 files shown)


RocketD0g:
Spot checking, the PM2.5 stuff looks good to me. From the original data sources, I am surprised at same BC values that are repeating for long periods of time. Will check in on this with SPARTAN folks before adding in BC, but I think we're good to add the PM2.5.

Before we add in any data, however, we'll need to address this Issue: https://github.com/openaq/openaq-data-format/issues/8

PR #183, opened 2016-08-02
WIP - EEA adapter

olafveerman:
Still to do:
- [x] determine the correct setup. One adapter, multiple sources?
- [x] find out how EEA is handling dates and TZ.
- [x] determine approach to parse location names


jflasher:
So for the one adapter, multiple sources question, what are others' thoughts? For AirNow, it's one adapter, one source even though there are different countries. The adapter just runs once and grabs it all.

I'd definitely only want one adapter, but we could do one source for each country, which would be more explicit. The whole source JSON object gets passed to the adapter so we could put something in the source `country` and check that in the adapter if we wanted.

olafveerman:
Pros of tying each source to a single country:

- it's clean and explicit
- no changes needed to the `/sources` endpoint
- provides a higher degree of control. If would want to pull data for Holland from the original source instead of the EEA, that's possible

The downside:

- the system won't automatically capture new countries added to a source
- more source files to manage

Anything else? I'm starting to like the idea of enforcing a source per country more and more.

jflasher:
Yeah, I think I'm liking source per country as well. Also, we can add a `custom: {}` property to the source.json file that has whatever extra data you want to put in it that'll get passed to the adapter if that's helpful at all.

jflasher:
Made the following updates:

- added location name
- properly setting date (hopefully)
- narrowed from/to dates
- added averaging/attribution

[at.txt](https://github.com/openaq/openaq-fetch/files/506742/at.txt)
[be.txt](https://github.com/openaq/openaq-fetch/files/506741/be.txt)

Added the two text files above to be used for spot checking for AT and BE, I think @RocketD0g was able to do this. Main thing we need to confirm is the date is handled properly. From the docs, I believe EEA is using local time for their dates.

@olafveerman could you look this over and see if it looks alright? I was looking it over and believe it's ok. For the location name, I just used station_name which looks to be as good as we'll get. And for city, we use network name which will probably sometimes definitely not be the city, but again, probably as close as we'll get.

jflasher:
👀 this over again and I'm going to pull it in. 🚀

PR #213, opened 2016-12-10
[HOLD] Switching to using EEA’s async requests

jflasher:
- Also upping request timeout to 3 minutes


jflasher:
Note that this should not be pulled in until a character encoding problem with EEA async system is fixed or worked around.

jflasher:
Superseded by https://github.com/openaq/openaq-fetch/pull/245

PR #232, opened 2017-01-06
Add Slovenian Environment Agency source

### LINKED ISSUE #61: New Data Source - Slovenia
Data submitted by someone from the Source Google form for Slovenia: 

Hourly: http://www.arso.gov.si/xml/zrak/ones_zrak_urni_podatki_zadnji.xml
Daily: http://www.arso.gov.si/xml/zrak/ones_zrak_dnevni_podatki_zadnji.xml"

jure:
This closes #61.

### DIFF
--- a/adapters/slovenia.js
+++ b/adapters/slovenia.js
@@ -1,0 +1,120 @@
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+import _ from 'lodash';
+import { default as moment } from 'moment-timezone';
+import cheerio from 'cheerio';
+
+exports.name = 'slovenia';
+
+exports.fetchData = function (source, cb) {
+  request(source.url, function (err, res, body) {
+    if (err || res.statusCode !== 200) {
+      return cb({message: 'Failure to load data url.'});
+    }
+
+    // Wrap everything in a try/catch in case something goes wrong
+    try {
+      // Format the data
+      var data = formatData(body, source);
+
+      // Make sure the data is valid
+      if (data === undefined) {
+        return cb({message: 'Failure to parse data.'});
+      }
+      cb(null, data);
+    } catch (e) {
+      return cb({message: 'Unknown adapter error.'});
+    }
+  });
+};
+
+var formatData = function (data, source) {
+  var getDate = function (dateString) {
+    var date = moment.tz(dateString, 'YYYY-MM-DD HH:mm', 'Europe/Ljubljana');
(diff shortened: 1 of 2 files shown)


sources/si.json
```diff
@@ -0,0 +1,16 @@
...
+        "url": "http://www.arso.gov.si/xml/zrak/ones_zrak_urni_podatki_zadnji.xml",
+        "adapter": "slovenia",
+        "name": "Slovenian Environment Agency",
+        "city": "",
+        "country": "SI",
+        "description": "Official hourly air pollution data by the Slovenian Environment Agency",
+        "sourceURL": "http://www.arso.gov.si",
+        "resolution": "1 hr",
+        "contacts": [
+            "gp.arso@gov.si"
```
jflasher: `contacts` is currently unused, but it's purpose is to be used to notify a group of people if there is a problem with the source (generally those who wrote the adapter). This looks like it might be the official SI gov email address? Can you also add info@openaq.org to the list?

adapters/slovenia.js
```diff
@@ -0,0 +1,120 @@
...
+
+      var unit = getUnit(this.tagName);
+      var value = parseFloat($(this).text());
+
+      if(unit === 'mg/m³') {
+        value = value * 1000
+      }
+
+      if (unit && value) {
+        obj.city = city
```
jflasher: Can you include `obj.location=` something meaningful in here? Otherwise it gets the source name which can lead to uniqueness issues when saving to the database.

adapters/slovenia.js
```diff
@@ -0,0 +1,120 @@
...
+    var city = $(elem).children('merilno_mesto').text();
+
+    $(elem).children().each(function (i, e) {
+
+      var obj = _.cloneDeep(baseObj);
+
+      var unit = getUnit(this.tagName);
+      var value = parseFloat($(this).text());
+
+      if(unit === 'mg/m³') {
```
jflasher: Can you use utils.convertUnits on final `measurements` array like in https://github.com/openaq/openaq-fetch/blob/develop/adapters/airnow.js#L160?

jflasher:
From #61, sounds like there may still be some missing lat/lon and pending discussion, we may want to add PM10 for other stations? Added some comments and otherwise, looking 👍 

Travis is failing because of syntax rules https://travis-ci.org/openaq/openaq-fetch#L1292

jure:
Thank you for the comments, all reasonable stuff. I'll fix it up and limit
the measurements to PM10 only and this one will be ready to go.

The Ljubljana source will be in a different PR, and that's the one I'm
missing lat/lon for.

jflasher:
Hi @jure, wanted to check in here and see if you were interested in making the changes or if we should open it up to the community? Thanks!

PR #240, opened 2017-02-05
Fixed minor typo

gaomrx:
"run" to "running"


jflasher:
Thanks for the PR @gaomrx. I think this reads fine as is, so going to close this, but glad to have some more eyes going over the documentation!

PR #266, opened 2017-02-20
Update DEFRA adapter with two location data

dolugen:
There were several failures for the latest fetch on status.openaq.org. One of them was of DEFRA:

```
DEFRA - instance requires property "city" (2 failed)
```
I looked into `defra.js` and saw that the city and coordinates of locations were hardcoded. I did some comparison between existing UK data in API and the live source. There were 2 new locations. And some missing too.

### DIFF
--- a/adapters/defra.js
+++ b/adapters/defra.js
@@ -439,7 +439,7 @@ let metadata = {
    { city: 'Belfast',
      coordinates: { latitude: 54.572586, longitude: -5.974944 } },
   'Derby St Alkmund\'s Way':
-   { city: 'Bass\'s Recreation Ground St Alkmund\'s Way Derby',
+   { city: 'Derby',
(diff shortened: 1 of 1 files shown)


adapters/defra.js
```diff
@@ -438,6 +438,9 @@ let metadata = {
   'Belfast Stockman\'s Lane':
    { city: 'Belfast',
      coordinates: { latitude: 54.572586, longitude: -5.974944 } },
+  'Derby St Alkmund\'s Way':
+   { city: 'Bass\'s Recreation Ground St Alkmund\'s Way Derby',
```
jflasher: The two cities here seem very specific, are they addresses maybe?

dolugen: They were taken from the results of `defra-locations.js` script. I quickly assumed that the results were used as is. I'll update with their city names: Derby and Plymouth respectively.

PR #296, opened 2017-04-27
Tuzla adapter

### LINKED ISSUE #276: Tuzla, Bosnia AQ data
We received a request while in Tuzla, Bosnia to add this data source:

http://monitoringzrakatk.info/bkc-today.html

If someone is interested in writing an adapter for this source and would like to also connect to the individual who recommended the source, let me know and I will connect you by email.

dolugen:
closes #276

### DIFF
--- a/adapters/tuzlanski.js
+++ b/adapters/tuzlanski.js
@@ -4,6 +4,7 @@ import { REQUEST_TIMEOUT } from '../lib/constants';
 import { default as baseRequest } from 'request';
 const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
 import cheerio from 'cheerio';
+import Coordinates from 'coordinate-parser';
 import { default as moment } from 'moment-timezone';
 import { flattenDeep } from 'lodash';
 import { parallel } from 'async';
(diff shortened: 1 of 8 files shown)


jflasher:
@dolugen Thanks! If you're all done, I can go ahead and review soon?

dolugen:
@jflasher Yes, please!

adapters/tuzlanski.js
```diff
@@ -0,0 +1,131 @@
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+import cheerio from 'cheerio';
+import Coordinates from 'coordinate-parser';
```
jflasher: What do you think about using http://turfjs.org/docs/#point for this? We're already using Turf in fetch (https://github.com/openaq/openaq-api/blob/develop/package.json#L41) and would be good to not add another library into the mix.

dolugen: I'm only using `coordinate-parser` to convert this format: `44°31´55´´N` to decimal degrees. I don't see how Turf does that.

dolugen:
> So much hackiness

I blame the markup.

adapters/tuzlanski.js
jflasher: Nope, you're right, wasn't really thinking over what the exact use case was.

PR #299, opened 2017-05-23
Adapter for German initiative luftdaten.info

lhufnagel:
Hey,
I wrote an adapter for the German initiative, see here https://openaq.slack.com/archives/C098H3JH2/p1494257255432403 

I'll message them and clarify few open questions:
- can we use the data (want to be a good user ;))
- is averaging period of 5 minutes true? (Api doc simply states that the call returns measurements from the last five minutes)
- how to map from sensor id to city/approximate location?

### DIFF
--- a/adapters/germany.js
+++ b/adapters/germany.js
@@ -1,0 +1,97 @@
+/**
+ * This code is responsible for implementing all methods related to fetching
+ * and returning data for the German data sources.
+ */
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT}); // might be to low, but I have bad wifi on a train, TODO: clarify
+import { default as moment } from 'moment-timezone';
+
+exports.name = 'germany';
+
+/**
+ * Fetches the data for a given source and returns an appropriate object
+ * @param {object} source A valid source object
+ * @param {function} cb A callback of the form cb(err, data)
+ */
+exports.fetchData = function (source, cb) {
+  request({url: source.url, headers: {'User-Agent': 'OpenAQ'}},
+  function (err, res, body) {
+    if (err || res.statusCode !== 200) {
+      return cb({message: 'Failure to load data url.'});
+    }
+
+    // Wrap everything in a try/catch in case something goes wrong
+    try {
+      // Format the data
+      const data = formatData(body);
+
+      // Make sure the data is valid
+      if (data === undefined) {
+        return cb({message: 'Failure to parse data.'});
+      }
+      cb(null, data);
+    } catch (e) {
+      return cb({message: e});
+    }
+  });
+};
+
+/**
+ * Given fetched data, turn it into a format our system can use.
+ * @param {array} results Fetched source data and other metadata
+ * @return {object} Parsed and standarized data our system can use
+ */
+const formatData = function (data) {
+  // Wrap the JSON.parse() in a try/catch in case it fails
+  try {
+    data = JSON.parse(data);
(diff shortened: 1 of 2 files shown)


dolugen:
I was a bit confused at first that Github showed me commits about the Sweden adapter.

lhufnagel:
Yes, I somehow included my old commits in the PR by accident. However, as discussed in Slack these personal sensors are currently unlikely to be included in the database

Max-Might:
Is there any progress on this? It would be nice to have data from this project.
There is something I would like to point out though. Luftdaten is a German project but it provides data for many more countries, not just Germany. Just check out the [map](http://maps.luftdaten.info/#2/0.0/0.0).

The idea is that you can buy a sensor and you can install it at your place and configure it to send measurements to their server. Current sensors measure P2.5 and P10 but more are going to be added in the future as far as I understand.

lhufnagel:
Have you seen this https://openaq.slack.com/archives/C0Q5NB0NN/p1495730356303957

Max-Might:
No, I don't use slack and this page is not public (requires login).

jflasher:
@Max-Might pasting the referenced comments below.

christa
[12:39 PM] 
Hey lhufnagel - This is awesome, thanks! And also awesome for looking into those details on data usage. To confirm with you (or perhaps dirk?): Are these DIY? Or are these a) government run air quality sensors or b) a research grade work (e.g. associated with a university or research institution and intended for publication)? We're definitely interested in sharing 'other' (e.g. DIY/low-cost sensor data from the public) data, but our system can't yet include that information. We're seeking funding to expand to that (as well as adjust our system to adequately share/communicate/visualize the difference in data types), but can't properly support it yet.



lhufnagel [12:50 PM] 
Hey christa alright, these are DIY/low-cost-sensor data. Good to know that you're looking into expanding the system for that :slightly_smiling_face:

jflasher:
I'm going to close this issue for now so it doesn't sit around and make me feel guilty. :) But we can reopen it once we're at a point to accept these types of sources. Thanks @lhufnagel !

RocketD0g:
Hey @lhufnagel, Thanks again for writing this adapter. I'm scoping out the feasibility of our system ingesting some low-cost sensor data in some manner as a pilot, and this data could be great place to start. 

Wondering: How many stations are there that this adapter would add in - and do you know if there is any possibility of accessing the data at longer averaging intervals than every 5 min? Say hourly? 

Also tagging @dirkschumacher since I know he is running one in Berlin, per the Slack convo.

lhufnagel:
Hey Christa, thanks for following up! 
As of right now, I'm counting approx. 3k sensors. However, it's my understanding that the API only provides data of these that have been active in the last five minutes (and there were ~7k measurements, of different quantities); five minutes is also how long measurement data is averaged. I am just gathering that info from their FAQ - I have never heard back from them on my original message in May; I'll resend and inquire about longer averaging periods and keep you updated :)

lhufnagel:
No reply either :(

dirkschumacher:
Who did you try to contact?

lhufnagel:
http://luftdaten.info/kontakt/ I used their contact form

dirkschumacher:
I talked to Rajko (http://luftdaten.info/kontakt/) a couple of times and he was very responsive.

lhufnagel:
So I got a reply, stating that they are currently having internal discussions about overall quality of the measurements and would not really recommend spreading the data right now.
They have data-streams with different averaging-intervals, but that should be dealt with, once quality issues are sorted out. And they have nearest-city-information in their database for some sensors, but not all and in no unified format.
So, hm...

RocketD0g:
Thanks for checking in with them, @lhufnagel, and this update!

PR #301, opened 2017-05-31
Moscow adapter

dolugen:
* Assuming averagingPeriod as 1 hour
* ~~TODO: Add coordinates~~
* ~~TODO: Exclude mobile stations~~

Ref: #7 (<--that is old)


dolugen:
It's missing the coordinates, but other than that, should be fine. Can you take a look at this @jflasher ?

jflasher:
@dolugen I think we've got a lead on the coordinates? If so, can we hold off to review until then? It'll keep us from having to add them after the fact if we pull in data without them.

dolugen:
Ok, then I'll ping you once I finish with the coordinates.

dolugen:
@jflasher So I've got the coordinates in the adapter now, like:

```javascript
"Кожуховский проезд": {
		"coordinates": {
			"longitude": 37.6585581,
			"latitude": 55.7087026
		}
	},
"Шаболовка": {
		"coordinates": {
			"longitude": 37.6066741,
			"latitude": 55.726868
		}
	},
```

The problem is, I can't read the coordinates of stations with names containing space (or dots, parens, etc.). `locations["Кожуховский проезд"]` -> no result. But, `locations["Шаболовка"]`is found OK. Any idea what the issue is here?

jflasher:
@dolugen I wonder if it might be a unicode error? Are the spaces somehow encoded differently?

dolugen:
I'm just gonna step over that problem and use the unique ID in the station URL as key. The mobile check is a bit late, but it was the quickest thing I could do.

You can review it now, thanks.

adapters/moscow.js
```diff
@@ -0,0 +1,238 @@
...
+
+import { default as baseRequest } from 'request';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+import cheerio from 'cheerio';
+import { default as moment } from 'moment-timezone';
+var Iconv = require('iconv').Iconv;
+import { flattenDeep, isFinite } from 'lodash';
+import { parallel } from 'async';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
```
jflasher: Can you move this above line 4 where it's used?

adapters/moscow.js
```diff
@@ -0,0 +1,238 @@
...
+      }
+    }
+
+    for (let i in parameters) {
+      if (parameters.hasOwnProperty(i)) {
+        let m = Object.assign({}, base);
+        m.parameter = parameters[i].name;
+        m.unit = parameters[i].unit;
+        if (m.unit === 'мг / куб. м') {
+          m.unit = 'mg/m³';
```
jflasher: Are we sure it's milli and not micro? @RocketD0g is confirming as well, but just want to make sure we get that right.

RocketD0g: @dolugen's correct that it is milli, not micro.

dolugen: Pretty sure:

* [Миллигра́мм (русское обозначение: мг, международное обозначение: mg)](https://ru.wikipedia.org/wiki/%D0%9C%D0%B8%D0%BB%D0%BB%D0%B8%D0%B3%D1%80%D0%B0%D0%BC%D0%BC)
* [микрограмм (мкг или µg)](https://ru.wikipedia.org/wiki/%D0%9C%D0%B8%D0%BA%D1%80%D0%BE%D0%B3%D1%80%D0%B0%D0%BC%D0%BC)

adapters/moscow.js
dolugen: Ok.

dolugen:
Thanks for the review, I'll take a look at the odd date issue and get back to you.

dolugen:
I was using the wrong method for the day of the month! Thanks for catching that.

jflasher:
Looks good, squashing and pulling in! Thanks @dolugen!

Issue #310, opened 2017-06-29
Update moment-timezone version


dolugen:
It's important to keep up to date with timezone changes. As an example, [Mongolia stopped observing DST since last February](http://mm.icann.org/pipermail/tz-announce/2017-February/000045.html). Despite IANA timezone database updates, older systems are still using the ULAST+0900 timezone. This is true for `openaq-fetch`'s `moment-timezone` version. It should be updated to the latest `2017b` version.

(I wanted to confirm `Agaar.mn`'s output for this, but it keeps timing out. `StateAir_Ulaanbaatar` confirms this.)

Now this at least affects Mongolian AQ data since February, which I'll be creating a separate issue for. Any other affected countries should be investigated and confirmed.

dolugen:
This blocks #311.

dolugen:
moment-timezone changelog, if interested: https://github.com/moment/moment-timezone/blob/master/changelog.md

jflasher:
Fixed with https://github.com/openaq/openaq-fetch/pull/312

PR #312, opened 2017-07-02
Update moment-timezone version in package.json #310

dolugen:


dolugen:
@jflasher `docker-compose` isn't behaving as usual.

dolugen:
All good now.

Issue #311, opened 2017-06-29
Fix MN timezone in measurements since Feb 2017


dolugen:
As described in #310, MN timezones should be updated from +0900 to +0800 for affected data points.

dolugen:
I believe this is done.

PR #316, opened 2017-07-04
Add migration to fix Ulaanbaatar timezone

dolugen:
Related: #311.

This is a basically a minor adaptation of the Amsterdam migration.


jflasher:
Thanks! Going to figure out what caused https://github.com/openaq/openaq-api/issues/326 before merging this in.

PR #317, opened 2017-07-06
Migrations to fix AU, BR data

dolugen:
This should fix:
* openaq/openaq-api#301
* openaq/openaq-api#326

I'd run these on a test/staging DB with real data.


jflasher:
Thanks @dolugen! I'll find some time to check these out. Depending on how long they take, I may do them outside the normal fetch process as I don't want it to hold that up and then manually mark them as migrated.

majesticio:
I am closing this as it has become a bit stale, but we do appreciate your contributions and are open to new information and insights regarding this PR. Thanks!

PR #325, opened 2017-07-14
add buenos aires adapter

maxgrossman:
Adapter for Buenos Aires.


adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+import cheerio from 'cheerio';
+import async from 'async';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+
+export const name = 'buenos aires';
+export function fetchData (source, callback) {
+  // list of requests used to get data
+  const tasks = generateTasks(source);
+
+  async.parallel(tasks, (err, response) => {
```
dolugen: I bet Joe will ask you to import `parallel` explicitly ;-)

adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+  });
+}
+
+// makes a requests list of requests for each station to each pollutant
+const generateTasks = (source) => {
+  const linkRequests = [];
+  ['CENTENARIO', 'CORDOBA', 'LA BOCA'].forEach((station) => {
+    ['CO', 'NO2', 'PM10'].forEach((pollutant) => {
+      const makeLinks = (station, pollutant) => {
+        const day = moment().date().toString();
```
dolugen: I think `moment().format()` is better suited here.

```js
>> var d = moment();
>> d.format("D");
"15"
>> d.format("MM");
"07"
>> d.format("Y")
"2017"
```

Also, is `moment.tz` useful here?

adapters/buenos-aires.js
jflasher: Yes please!

adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+          case 'NO2':
+            pollutantNum = 2;
+            break;
+          case 'PM10':
+            pollutantNum = 3;
+            break;
+          default:
+            break;
+        }
+        const link = [
```
jflasher: Can you do this as a string template literal? This just reads as a bit confusing to me.

adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+    } catch (e) {
+      return callback(new Error('Unknown adapter error.'));
+    }
+  });
+}
+
+// makes a requests list of requests for each station to each pollutant
+const generateTasks = (source) => {
+  const linkRequests = [];
+  ['CENTENARIO', 'CORDOBA', 'LA BOCA'].forEach((station) => {
```
jflasher: Can we make this an array of objects like [{name: 'Centenario', number: 1, coordinates: {lat: 23, lon: 23}}, ...] so that we can avoid so many switch statements later? 

Also, is there any way to get this info programatically instead of hard coding it?

adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+    ['CO', 'NO2', 'PM10'].forEach((pollutant) => {
```
jflasher: Same as comment above but include pollutant name, unit and period?

adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+          '&fecha_anio=',
+          year,
+          '&menu_id=34234&buscar=Buscar'
+        ].join('');
+        const linkRequest = (callback) => {
+          request.get({
+            url: link
+          }, (err, res, body) => {
+            if (err || res.statusCode !== 200) {
+              return callback(new Error('was not able to retrieve' + pollutant + ' data for ' + station));
```
jflasher: So in this case, I believe if 1 url request is unavailable, all of them will be marked as failed. Is this what we want?

adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+
+// returns list of measurements generated from requests.
+const formatData = (htmlList) => {
+  let measurements = htmlList.map((html) => {
+    // data is in imgs' title tags like [pollutantVal - hr]
+    // aqData is list of lists with said data
+    let aqData = html.match(/<img[^>]+>/g).filter((img) => {
+      return img.match(/hs/);
+    });
+    aqData = aqData.map((img) => {
```
jflasher: If aqData has no match above, I think this will crash?

adapters/buenos-aires.js
```diff
@@ -0,0 +1,192 @@
...
+      'DD/MM/YYYY HH:mm:ss',
+      'America/Argentina/Buenos_Aires'
+    );
+    // get contaminante, pollutant, and estacion, station. use them in above set** functions
+    // when populating measurement
+    const $ = cheerio.load(html);
+    const contaminante = $('#contaminante').html().split('selected>')[1].split('<')[0];
+    const estacion = $('#estacion').html().split('selected>')[1].split('<')[0];
+    const measurement = {};
+    measurement['parameter'] = contaminante === 'PM10' ? 'pm10' : contaminante;
```
jflasher: This is totally personal preference, so feel free to ignore, but I try to write these like

```js
const measurements = {
  parameter: contaminante === 'PM10' ? 'pm10' : contaminante,
  date: {
    utc: aqData[1].toDate(),
    local: aqData[1].format()
  }
};
```
I feel it reads better, but like I said, totally personal preference.

adapters/buenos-aires.js
maxgrossman: Very much support this change on second look at the code. 

Is the hard coding concern in regards too much unnecessary code I wrote? Or is it that the assumed mapping between form value and request I'm making might change? If the latter is the worry I can grab the values from the form with a request and use those to make the links.

adapters/buenos-aires.js
maxgrossman: Changed and will be reflected in upcoming push.

adapters/buenos-aires.js
jflasher: Totally the latter. We'd like the adapters to be as dynamic as possible. So if a new location is added, ideally, it'd be automatically picked up by the adapter.

maxgrossman: Makes sense! Will make this change in my new push.

adapters/buenos-aires.js
```diff
@@ -0,0 +1,208 @@
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import { default as moment } from 'moment-timezone';
+import cheerio from 'cheerio';
+import { parallel } from 'async';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+
+export const name = 'buenos aires';
```
jflasher: Just a note that this doesn't match what's in the source.

sources/ar.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "url": "http://www.buenosaires.gob.ar/areas/med_ambiente/apra/calidad_amb/red_monitoreo/index.php?",
+        "adapter": "buenos-aires",
+        "name": "Buenos Aires",
+        "country": "AR",
```
jflasher: "active": true needs to be added here or else the source won't be run.

dolugen:
I've talked about Max about progressing some of the PRs and now working on to finish this adapter.

jflasher:
👍 Thanks @maxgrossman and @dolugen

Issue #146, opened 2016-04-12
Data from Israel


maelle:
The website is http://www.svivaaqm.net/Default.rtl.aspx

When playing with grey buttons on the right I e.g. got the two tables I'm attaching.

![israel](https://cloud.githubusercontent.com/assets/8360597/14472311/e0fc7aa4-00f2-11e6-9086-e41ce56f37c6.png)
![israel2](https://cloud.githubusercontent.com/assets/8360597/14472312/e118fbe8-00f2-11e6-8ffb-e15c0ab892f0.png)

masalmon:
"נתוני ניטור בזמן אמת" means "real time monitoring data"

masalmon:
and for instance "חיפה וקריות" is "Haifa and its suburbs"

masalmon:
Last thing for today, the column name on the right in the table is "station name".

dolugen:
Anyone know Hebrew?

masalmon:
Should we make a Twitter call for help?

jflasher:
If we leave the place names in Hebrew, I think we don't need to know any to grab the data? Only thing I don't see is lat/lon.

Does leaving the names in Hebrew sound good? Otherwise I guess we'd be hardcoding translations.

RocketD0g:
Sounds good to me!

masalmon:
A.k.a. "If you have encoding issues in your code, you won't miss them now" :grinning:

jflasher:
😁

masalmon:
I think that for each station this page http://www.svivaaqm.net/StationInfo5.aspx?ST_ID=11 gives lon and lat?

jflasher:
Nice! Any way to get from that main page to the station pages? Is there an ID or similar?

masalmon:
קו אורך is longitude.

you get the station ID when your mouse is on the station name

![israel](https://cloud.githubusercontent.com/assets/8360597/21145400/64dcb1a6-c14e-11e6-8e1c-1f2690b1435a.png)

masalmon:
(station name being the column in Hebrew)

masalmon:
On each station page clicking on the button in the red circle on this pic, you get to a data query page. I guess one could get historical data from there?

![station](https://cloud.githubusercontent.com/assets/8360597/21145459/b3e48292-c14e-11e6-96a6-3bf311b6a344.png)

masalmon:
[...]

PR #328, opened 2017-07-17
Adapters/israel

maxgrossman:
Add adapter for Israel.


jflasher:
Per https://github.com/openaq/openaq-fetch/issues/146#issuecomment-315729551, is this ready to be looked at yet? 

Also, can you add associated source json block so I can test it?

maxgrossman:
@jflasher yes this is ready to be looked at. 

Should have followed up with a comment friday afternoon, but after refactoring Dolugen's code for Israel, I still came across the same form submit problem in the end. The site backends are certainly the same, but the way they post the data on form request is different.

Anyhow, please let me know if you have questions with this code. Looking at it now I think I see a problem in right in the fetchData function. I will let you look into it but perhaps I need to just move [line 17](https://github.com/openaq/openaq-fetch/blob/adapters/israel/adapters/israel.js#L17 inside the try / catch statement and just call `handleState`?

maxgrossman:
@jflasher - after much collaboration with dolugen collaboration, this is ready for a review. 

three things to note

1) the `dynamicTableFallback` function, which is fired when the main method for grabbing data fails, fires conditional to the averaging period of the raw data being 10 min or greater (per the data requirements). 

In the dryruns, I noticed this made the fallback function for naught, as those stations that used the fallback has averaging periods of 5 minutes. Just wanted to let you know.

2) `dynamicTableFallback` only retrieves the most recent record (as it is the only record available on the page it scrapes), while the main scraping method gathers a day's worth of data each fetch. 

If a handful of stations consistently rely on `dynamicTableFallback` for data, this could mean the data reported for those stations v. those that use the main method are inconsistent.

3) Perhaps because we are limiting parallel processes or the this adapter's reach into many different parts of the site, the run on my local machine was 2 minutes and 40 seconds.

thanks!
cc @dolugen @kamicut

package.json
```diff
@@ -7,6 +7,7 @@
     "lint": "eslint .",
     "test": "npm run lint && mocha test/** --require babel-register",
     "start": "node index.js",
+    "start-test": "babel-node adapters/test_and.js",
```
dolugen: Should this be removed?

adapters/envista_il.js
```diff
@@ -0,0 +1,1573 @@
...
+      const tasks = [queryStation(source, link, form, j, minAvgPeriod)];
+      parallel(tasks, (err, results) => {
+        return done(err, results);
+      });
+    });
+  };
+};
+
+// do station query
+const queryStation = function (source, link, qform, jar, minAvgPeriod) {
```
dolugen: `jar` isn't used here, so you can remove it.

adapters/envista_il.js
```diff
@@ -0,0 +1,1573 @@
...
+    const $ = cheerio.load(data, { xmlMode: true });
+    let base = {
+      location: location,
+      averagingPeriod: {unit: 'hours', value: 0.25},
+      attribution: [{
+        name: 'Israel Ministry of Environmental Protection',
+        url: source.url
+      }]
+    };
+    matchCoordinates(location);
```
dolugen: I believe this shouldn't be here.

package.json
```diff
@@ -16,6 +17,7 @@
     "async": "^2.5",
     "babel-preset-es2015": "^6.3.13",
     "babel-register": "^6.3.13",
+    "babyparse": "^0.4.6",
```
dolugen: I think this isn't used, so please remove it.

sources/il.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+    "url": "http://www.svivaaqm.net/",
+    "adapter": "envista_il",
+    "name": "Israel",
+    "country": "IL",
+    "city": "FIXME",
```
dolugen: The adapter isn't setting any `city` fields, any idea how to figure out the cities?

maxgrossman: we could use the regions the stations are in as the `regional approximates` discussed in the data format repo

maxgrossman:
@dolugen I added the city from regions and got rid of the unneccessary things you caught. 

To the point of the fallback - If consensus is it doesn't make sense, let's totally drop it. 

My argument for keeping it (with no pride being the one who wrote it, I promise) is that keeping it does not make refactorin a new envista source more difficult, nor does it seem to gobble up time when compared to runs where I left it out.

I'd also say, the fallback sorta lives on its own and only run if a try/catch finds an error. Plus while the majority of raw data here is at 5min intervals, there's no saying a longer period sensor won't need the fallback for some extended period of time. 

tl,dr: There might be edge cases where the fallback makes sense, but if any of the other reviewers agree its unneccessary, let's drop it!

adapters/envista_il.js
```diff
@@ -0,0 +1,1572 @@
...
+      form['BasicDatePicker2$textBox'] = $('#BasicDatePicker2_TextBox').attr('value');
+      form['txtEndTime'] = '00:00';
+      form['txtEndTime_p'] = txtEndTime.format('YYYY-M-DD') + '-0-0-0-0';
+      form['txtStartTime_p'] = txtEndTime.subtract(1, 'day').format('YYYY-M-DD') + '-0-0-0-0';
+      form['ddlAvgType'] = 'AVG';
+      form['ddlTimeBase'] = 15;
+      form['btnGenerateReport'] = 'הצג+דוח';
+
+      // used to ignore fallbacks w/ 5 min avaraging period
+      const minAvgPeriod = parseInt($('#ddlTimeBase').children()[0].attribs.value);
```
jflasher: What's this being used for? The averagingPeriod should be a property of the source and independent of when we're checking. Like, the data should be updating on the source every 10 minutes, no matter how often we're checking. Is that getting captured here?

adapters/envista_il.js
```diff
@@ -0,0 +1,1572 @@
...
+        return acceptableParameters.indexOf(match[1].toLowerCase()) >= 0;
+      }).each(function (i, el) {
+        let paramUnitM = rParamUnit.exec($(this).text());
+        let m = Object.assign({}, base);
+        if (m !== null) {
+          m.parameter = paramUnitM[1].toLowerCase();
+          m.date = dateProp;
+          m.coordinates = matchCoordinates(location);
+          m.unit = paramUnitM[2];
+          m.value = Number($(this).next().text());
```
jflasher: I saw a bunch of 0's in the dryrun results. I just want to make sure those are actual 0's coming from the source and not what we're putting in when there is no measurement coming from the source?

jflasher:
@dolugen @maxgrossman if we think the fallback will provide value, I'm all for keeping it. If we think it's just going to make the code harder to read for a minimal amount of extra data, I'd be in favor of keeping the code simpler so someone else can more easily fix anything in the future.

adapters/envista_il.js
maxgrossman: What's getting captured here is the minimum averaging period provided by the source for a given station. On the site you are able to select from multiple averaging periods for a given station. 

This min averaging period matches the averaging period reported on the page I was gathering the station fallback data. So the idea was if the averaging period for a given station was less than 10 minutes, the adapter would ignore trying to call the fallback.

dolugen:
I updated the Envista adapter to handle both sources in #358. I continued from my existing branch, so created a new PR for it.

PR #329, opened 2017-07-17
Fix header formatting

eliLitwack:


jflasher:
💥 making it on the contributor's list!

Issue #246, opened 2017-02-10
Andalucia data


maelle:
@Pakillo just told me about https://github.com/SevillaR/aire but it's written in R.

jflasher:
It does point to http://www.juntadeandalucia.es/medioambiente/site/portalweb/menuitem.7e1cf46ddf59bb227a9ebe205510e1ca/?vgnextoid=7e612e07c3dc4010VgnVCM1000000624e50aRCRD&vgnextchannel=3b43de552afae310VgnVCM2000000624e50aRCRD, but I can't seem to find the underlying data.

Pakillo:
Hi, 

All data are structured as daily reports. To see the data you need to (manually) select the province, date, and type of report (quantitative) and then you get HTML with data tables, like for example http://www.juntadeandalucia.es/medioambiente/atmosfera/informes_siva/feb17/nal170209.htm

jflasher:
Thanks @Pakillo! The URL format seems easy to break down. It's got the month + year and then a unique station code and date. So we should programmatically be able to build that URL for all the stations and skip the click the menu items part.

And a couple of quick questions.

1. Do you know if there are coordinates for the stations provided anywhere?
2. Do you know what the **PART** measurements are? I'm guess they're pm25 or pm10, but not sure which one.
3. Just to confirm this isn't in the EEA data? cc/ @masalmon

masalmon:
@jflasher I had wondered this and when looking at the grey points on OpenAQ map the South of Spain doesn't have any point at all so I don't think it was included in the EEA data?

jflasher:
Thanks @masalmon. And hopefully those points won't be grey for much longer!

masalmon:
Good thing Andalucia is one of the rare Spain regions I know where to place on a map :grin:

Pakillo:
Hi,

Glad you'll be able to integrate these into OpenAQ :) Note they are **raw** data, not validated, as captured by measuring stations. The Andalusian government supposedly passes some checks/quality control on them later. But I haven't been able to find the validated anywhere (which doesn't mean they wouldn't give them if we asked! But that's not automatic of course).

Note also that daily reports, and hence URLs, are by **province**, not station (e.g. al for Almeria, se for Sevilla, etc). Hence the same URL contains data for all the stations in that province (as separate HTML tables).

To your questions:

[...]

PR #330, opened 2017-07-18
Adapter/andalucia

maxgrossman:
Add an adapter for Andalucia data source.


sources/an.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "url": "http://www.juntadeandalucia.es/medioambiente/site/portalweb/menuitem.7e1cf46ddf59bb227a9ebe205510e1ca/?vgnextoid=7e612e07c3dc4010VgnVCM1000000624e50aRCRD&vgnextchannel=3b43de552afae310VgnVCM2000000624e50aRCRD",
+    "adapter": "andalucia",
```
jflasher: Can you give this a `name` so it can be run to test?

adapters/andalucia.js
```diff
@@ -0,0 +1,925 @@
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import cheerio from 'cheerio';
+import _ from 'lodash';
```
jflasher: Please import only methods needed.

adapters/andalucia.js
```diff
@@ -0,0 +1,925 @@
...
+import { default as moment } from 'moment-timezone';
+import async from 'async';
```
jflasher: Same as lodash, please import only methods needed.

adapters/andalucia.js
```diff
@@ -0,0 +1,925 @@
...
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+
+exports.name = 'andalucia';
+
+export function fetchData (source, callback) {
```
jflasher: Can we get some try/catch in here like previous adapters have in case there are issues running the code block?

adapters/andalucia.js
```diff
@@ -0,0 +1,925 @@
...
+  return (done) => {
+    // Try getting data from today and yesterday
+      // in both cases, if unsuccessful, return blank list
+      // when successful, return list of parsed data.
+    async.parallel([
+      (cb) => {
+        request.get(
+          url, (err, res, body) => {
+            if (err || res.statusCode !== 200) {
+              console.log(
```
jflasher: No console logging please for production.

adapters/andalucia.js
```diff
@@ -0,0 +1,925 @@
...
+      }
+      res = res[0].concat(res[1]);
+      done(null, res);
+    });
+  };
+};
+
+const formatData = ($, stations) => {
+  const tables = [];
+  // the tables we want don't have cellpadding 5.
```
jflasher: Is there any better way to do this? If it's the only way, so be it, but just feels so breakable. 😭

adapters/andalucia.js
```diff
@@ -0,0 +1,925 @@
...
+      tds.push($(el).text());
+    });
+    return tds;
+  });
+};
+
+const makeMeasurements = (stationData, stationLoc) => {
+  // get index of data sources we can record
+  const pollutantIndexes = stationData[0].filter((cell) => {
+    return _.includes(['SO2', 'NO2'], cell);
```
jflasher: Can you use the utils function acceptableParameters here to make it less hard coded?

adapters/andalucia.js
```diff
@@ -0,0 +1,925 @@
...
+      if (cell !== undefined) {
+        return cell;
+      }
+    });
+  });
+  // merge all measurement lists into one
+  return [].concat.apply([], stationData);
+};
+
+const makeDate = (date) => {
```
jflasher: If I'm understanding this correctly, all of this is to get it in a format `moment` can handle? `moment` can be given any arbitrary string pattern, you just need to tell it what it is. So I think this could be greatly simplified.

adapters/andalucia.js
maxgrossman: Agreed it is less than awesome way - the HTML provides zero id/class to latch onto to make selections, unfortunately. 

Maybe instead of relying on HTML attributes I can use the station name (since I have it in the station constant) and select the table that includes the station name + the next table.

That would get us the 2 table groups we need.

dolugen:
Linking original issue for reference: #246

maxgrossman:
@jflasher this adapter is ready for another check over. Below is a snippet of my dry run results (which are very helpful!)

```
2017-07-28T14:51:21.077Z - info: ///////
2017-07-28T14:51:21.077Z - info: [Dry Run] New measurements inserted for Andalucia: 21650
2017-07-28T14:51:21.077Z - info: ///////
2017-07-28T14:51:21.077Z - info: Dryrun completed, have a good day!
```

adapters/andalucia.js
```diff
@@ -0,0 +1,920 @@
...
+      }
+    });
+  });
+  // make each row into a list of measurements
+  stationData = stationData.slice(1, -1).map((row, index) => {
+    return row.map((cell, index) => {
+      if (index !== 0) {
+        // ignore blank spaces
+        if (/\S/.test(cell)) {
+          console.log(cell);
```
jflasher: No console.log please.

adapters/andalucia.js
```diff
@@ -0,0 +1,920 @@
...
+          return {
+            city: stationName.replace(/\w\S*/g, (t) => { return t.charAt(0).toUpperCase() + t.substr(1).toLowerCase(); }),
```
jflasher: Sample output looks like 

```json
{"date":{"utc":"2017-07-27T22:00:00.000Z","local":"2017-07-28T00:00:00+02:00"},"parameter":"no2","location":"unused","value":9,"unit":"µg/m³","city":"Cobre Las Cruces","attribution":[{"name":"Ministry of Environment and Spatial Planning","url":"http://www.juntadeandalucia.es/medioambiente/site/portalweb"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":37.5550190832439,"longitude":-6.03616087321919},"country":"ES","sourceName":"Andalucia","sourceType":"government","mobile":false}
```

Looks like we're missing `location`, everything should have a unique location within a `city`.

adapters/andalucia.js
maxgrossman: blah, sorry thought I dropped that.

jflasher:
👍 💥 closes #246

Issue #282, opened 2017-04-06
Italy Data - ARPALAZIO


FabioMD1972:
http://www.arpalazio.net/main/aria/sci/annoincorso/chimici.php

This site is only in italian language,  I can help to translalate it.

RocketD0g:
Thanks, @FabioMD1972! 👍

jobonaf:
the URL for the hourly data for the whole Latio region (not only Rome) can be built as follows:
`http://www.arpalazio.net/main/aria/sci/annoincorso/chimici/<province>/DatiOrari/<province>_<pollutant>_<year>.txt`
with provinces: CC,FR,LT,RI,RM,VT
and pollutants: BENZENE,CO,NO2,NOX,NO,O3,PM10,PM2.5,SO2

the TXT file is a table with columns: julian day, hour and one column for each station

stations' metadata:
https://github.com/jobonaf/calicantus/blob/master/data/sites-info/metadata.ARPA-Lazio.csv

RocketD0g:
I'm having issues accessing the hourly data via a link like:

http://www.arpalazio.net/main/aria/sci/annoincorso/chimici/%3Cprovince%3E/DatiOrari/RM_SO2_2017.txt

Thanks for any thoughts on where I am going wrong, @jobonaf!

jobonaf:
You should substitute `<province>`with `RM`not only in the file name, but also in the path, like this: [http://www.arpalazio.net/main/aria/sci/annoincorso/chimici/RM/DatiOrari/RM_SO2_2017.txt](http://www.arpalazio.net/main/aria/sci/annoincorso/chimici/RM/DatiOrari/RM_SO2_2017.txt)

RocketD0g:
Ah yup, oops - Thanks! Now labelling `ready for dev`.

dolugen:
Thank you, @FabioMD1972 and @jobonaf!

Also, the PDF linked from the page has detailed information on the file format. [Link to Google Translated version.](https://translate.google.com/translate?hl=en&sl=it&tl=en&u=http%3A%2F%2Fwww.arpalazio.net%2Fmain%2Faria%2Fsci%2Fannoincorso%2FLegendaDatiChimici.pdf)

dolugen:
> stations' metadata:
> https://github.com/jobonaf/calicantus/blob/master/data/sites-info/metadata.ARPA-Lazio.csv

@jobonaf This is very helpful. Can I ask how you generated this data?

[...]

PR #333, opened 2017-07-22
Arpalazio adapter

dolugen:
#282


adapters/arpalazio.js
```diff
@@ -0,0 +1,298 @@
...
+
+exports.name = 'arpalazio';
+
+const baseUrl = 'http://www.arpalazio.net/main/aria/sci/annoincorso/';
+const provinceQueryPath = 'chimici/chimici.php';
+const hourlyAvgParam = 0;
+const dailyAvgParam = 3;
+const hourlyAvgPeriod = {unit: 'hours', value: 1};
+const dailyAvgPeriod = {unit: 'hours', value: 24};
+const dailyParameters = ['pm25', 'pm10'];
```
jflasher: Can I get some clarification on this daily vs hourly split? It seems like we should want hourly whenever possible? Do a quick spot check, I saw pm25 and pm10 available for Viterbo it looks like we're still grabbing the daily values?

sources/it.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+    "url": "http://www.arpalazio.net/main/aria/sci/annoincorso/chimici.php",
+    "adapter": "arpalazio",
+    "name": "ARPALAZIO",
```
dolugen: Gotcha.

adapters/arpalazio.js
dolugen: We've talked about it on #282 (starting at [comment](https://github.com/openaq/openaq-fetch/issues/282#issuecomment-317225793)).

sources/it.json
dolugen: They're using `ARPALAZIO`, an abbreviation of "Agenzia Regionale Protezione Ambientale Del Lazio".

dolugen:
> Also, can we put something in place that'll only grab data from a day or 2 back? Right now, it's trying to insert ~600,000 records which definitely doesn't seem necessary.

Do you want this in another branch? I was thinking you'd let this version run at least once, and then apply the recent-only change.

adapters/arpalazio.js
jflasher: Oh yeah, I remember that coming up now. This approach looks good.

jflasher:
@dolugen yeah, could you do that in another branch off this one? It's a good point about getting all the back data in place. We'll see what happens when it tries to insert 500,000 records :)

Once that other PR is in place, I'll merge this one and let it populate old data.

Issue #298, opened 2017-05-22
Investigate EEA's new direct download option


jflasher:
EEA rolled out a new access option that may be more reliable, check it out. 

More info to come.

lhufnagel:
Email from May 18th

Dear users of the EEA Up-to-date air quality download service
In the last two years we have seen a quickly growing interest in access to the up-to-date air quality download service. We are of cause very happy for the huge interest, but it has also pushed our services to the limit and even though we have been upgrading and improving the platform several times in the last year, we have not been able to keep up with the demand.

To provide a better service, we have therefore developed a simplistic direct download solution which we believe will cover the majority of downloads and hopefully also your need.

The direct download solution is now available on the UTD download page: 
http://discomap.eea.europa.eu/map/fme/AirQualityUTDExport.htm

Every 12 hours we extract the last 48 hours of data per country and the 8 major pollutants. Not all countries delivers all of these 8 pollutants so in total it is around 180 files. The files are only available in the CSV2-format which is known from the on-demand download service.
The file names are static, making it easy for your automated downloads to access the latest data. 
Sample url (Andorra, Ozone): http://discomap.eea.europa.eu/map/fme/latest/AD_O3.csv
Generic url: http://discomap.eea.europa.eu/map/fme/latest/<countrycode>_<pollutant>.csv
The data are provided in UTC+1 (the timezone is included in CSV2) and as with the on-deman service, only hourly data are provided.

We have decided only to offer the last 48 hours of data with this option. This is due to the fact, that we receive updates to already delivered data, and in order to keep the files in sync with our database we do not want to keep the old and possible out-of-sync files.

The old on-demand download service will continue to be supported, but we hope many of your will find the direct download useful and hereby removing the pressure on the on-demand download service.

Should you have any questions to the new option, please do not hesitate to contact me.

Kind regards
Peter Kjeld

maxgrossman:
@jflasher @lhufnagel - beginning work on this [here](https://github.com/openaq/openaq-fetch/tree/adapter/direct-eea). 

I'll use links like the one above in each country's source. Coords will come from [this](http://discomap.eea.europa.eu/map/fme/metadata/PanEuropean_metadata.csv) metadata file that, among other information, has EPSG:4979 coordinates for stations. 

I'll do this first for countries in #309 and push often!

jflasher:
[...]

PR #336, opened 2017-07-26
Adapter/direct eea

maxgrossman:


maxgrossman:
'ready for check'

adapters/eea-direct.js
```diff
@@ -0,0 +1,177 @@
+'use strict';
+import { acceptableParameters, convertUnits } from '../lib/utils';
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import { default as moment } from 'moment-timezone';
+import { parallel, map, filter } from 'async';
+import { parse } from 'babyparse';
+import uniqBy from 'lodash.uniqBy';
```
olafveerman: Discussing this with @maxgrossman, this was added in Lodash 4. Currently checking if upgrading lodash will break anything. Separate PR forthcoming.

adapters/eea-direct.js
```diff
@@ -0,0 +1,177 @@
...
+  }, (err, mappedRecords) => {
+    if (err) {
+      return cb(null, []);
+    }
+    cb(null, mappedRecords);
+  });
+};
+
+const makeCoordinates = (coordinatesList, stationId) => {
+  return coordinatesList.filter((coordinates) => {
```
olafveerman: Use `.find` instead of `.filter`

adapters/eea-direct.js
```diff
@@ -0,0 +1,177 @@
...
+    return coordinates.stationId === stationId;
+  }).map((station) => {
+    return {
+      latitude: parseFloat(station.coordinates.latitude),
+      longitude: parseFloat(station.coordinates.longitude)
+    };
+  })[0];
+};
+
+const makeAvgPeriod = (delta) => {
```
olafveerman: I was under the impression that the Averaging Period for EEA was always 1 hour.

adapters/eea-direct.js
```diff
@@ -0,0 +1,177 @@
...
+  const latestTime = moment.tz(delta[1], 'YYYY-MM-DD hh:mm:ss', 'Europe/Berlin');
+  const earliestTime = moment.tz(delta[0], 'YYYY-MM-DD hh:mm:ss', 'Europe/Berlin');
+  return moment(latestTime).diff(earliestTime, 'hours');
+};
+
+const makeDate = (date) => {
+  let timeZone = date.slice(date.length - 6, date.length);
+  switch (timeZone) {
```
olafveerman: Did some investigation around how EEA handles time. See the comment here: https://github.com/openaq/openaq-fetch/issues/298#issuecomment-318382840

adapters/eea-direct.js
```diff
@@ -0,0 +1,177 @@
+'use strict';
+import { acceptableParameters, convertUnits } from '../lib/utils';
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import { default as moment } from 'moment-timezone';
+import { parallel, map, filter } from 'async';
+import { parse } from 'babyparse';
```
olafveerman: Actually, we are already using `csv-parse` elsewhere. Is there a good reason to use `babyparse` instead?

maxgrossman:
@olafveerman - this is ready for a check.

maxgrossman:
@jflasher and @dolugen - the eea adapter is now ready for a check w/new station city/location names now coming from [battuta](https://github.com/openaq/battuta/tree/develop).

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
+'use strict';
+import { acceptableParameters, convertUnits } from '../lib/utils';
```
jflasher: Please add a line after `'use strict';`, my only reasoning for this is because I'm overly uptight about some things :)

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import { default as moment } from 'moment-timezone';
+import { parallel, map, filter } from 'async';
+import { default as parse } from 'csv-parse/lib/sync';
+import uniqBy from 'lodash.uniqby';
+import { bboxPolygon, inside, point } from 'turf';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+const stationsLink = 'https://raw.githubusercontent.com/openaq/battuta/develop/eea-stations.json';
```
jflasher: We should update this to /master/ or whatever the final branch for data is?

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+export const name = 'eea-direct';
+
+export function fetchData (source, cb) {
+  const metadataRequest = makeMetadataRequest(source);
+  const requestTasks = makeTaskRequests(source);
+  parallel(
+    [metadataRequest, requestTasks],
+    (err, res) => {
+      if (err) {
+        return cb(null, []);
```
jflasher: Please return nicer error message here.

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+    },
+    (done) => {
+      // try to combine locations with coordinates
+      // by matching stationId or a spatial join.
+      // then return successful combos
+      request.get({url: stationsLink}, (err, res, locations) => {
+        if (err) {
+          return done(null, []);
+        }
+        locations = JSON.parse(locations);
```
jflasher: Wrap in a try/catch please. JSON.parse's going bad have totally destroyed me before. :sob:

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+        // apply unit conversion to generated record
+        cb(null, convertUnits([measurement])[0]);
+      }, (err, mappedRecords) => {
+        if (err) {
+          return done(null, []);
+        }
+        done(null, mappedRecords);
+      });
+    },
+    (done) => {
```
jflasher: Just to confirm, this will only do the get on the stationsLike once correct?

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+          // try to spatial join
+          if (!(geocodedProps)) {
+            geocodedProps = locations.find((location) => {
+              if (location.stationId) {
+                const bbox = bboxPolygon(location.bounds);
+                const stationPoint = point(
+                  crds.longitude,
+                  crds.latitude
+                );
+                return inside(stationPoint, bbox);
```
jflasher: Can this return more than 1 element? Does that break anything?

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+  ], (err, mappedData) => {
+    if (err) {
+      return cb(err, []);
+    }
+    let measurements = mappedData[0];
+    const coordinates = mappedData[1];
+    // map coordinates and location names to measurements
+    map(measurements, (measurement, done) => {
+      let station = matchStation(coordinates, measurement['coordinates']);
+      measurement['location'] = (station.location ? station.location : 'unused');
```
jflasher: Please don't use `unused`, if we have no valid `location` or `city` name, just leave it undefined so it'll fail an insert or better yet, just exclude it from the list to be saved.

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+  }
+};
+
+const makeCoordinates = (coordinatesList, stationId) => {
+  return coordinatesList.find((coordinates) => {
+    return coordinates.stationId === stationId;
+  }).coordinates;
+};
+
+const makeDate = (date, timeZone) => {
```
jflasher: This seems overly complicated. We're gettign a timezone offset, and then tying it to a country's timezone, presumed to be in that area. But country timezones change at different times throughout the year. I think we should be directly using the data that they're given us, not proxying to some other country. 

Take a look at what the current EEA adapter does, I think the same should work here? https://github.com/openaq/openaq-fetch/blob/develop/adapters/eea.js#L94-L101

package.json
```diff
@@ -25,9 +25,10 @@
     "ftp": "^0.3.10",
     "iconv": "^2.2.3",
     "jsonschema": "^1.0.3",
-    "knex": "^0.13.0",
-    "knex-postgis": "^0.2.2",
-    "lodash": "^4.17.4",
+    "knex": "^0.9.0",
```
jflasher: I think these version downgrades got included by accident?

sources/be.json
```diff
@@ -1,7 +1,8 @@
+
 [
     {
-        "url": null,
-        "adapter": "eea",
+        "url": "http://discomap.eea.europa.eu/map/fme/latest/BE_<pollutant>.csv",
```
jflasher: I sorta want this to be 

```
"url": ""http://discomap.eea.europa.eu/map/fme/latest/"
```

and have the rest taken care of in the adapter, but I don't know why. I'm fine with this if you like it this way.

adapters/eea-direct.js
```diff
@@ -0,0 +1,265 @@
...
+      break;
+    case 'UTC-03':
+      timeZone = 'Atlantic/Bermuda';
+      date = date + '-03:00';
+      break;
+    default:
+      break;
+  }
+  date = moment.tz(date, 'YYYY-MM-DD hh:mm:ss', timeZone);
+  if (timeZone === 'Atlantic/Azores') {
```
jflasher: We should avoid this at all costs. :)

adapters/eea-direct.js
maxgrossman: yes this will only run one time in the adapter.

dolugen:
My notes:

* `turf` is missing
* Shouldn't all of the EEA sources be using `eea-direct`? I only see a few of them has it.
```
grep eea-direct sources/*
sources/ad.json:        "adapter": "eea-direct",
sources/be.json:        "adapter": "eea-direct",
sources/de.json:        "adapter": "eea-direct",
sources/fr.json:        "adapter": "eea-direct",

grep EEA sources/*
sources/ad.json:        "name": "EEA Andora",
sources/at.json:        "name": "EEA Austria",
sources/be.json:        "name": "EEA Belgium",
sources/cz.json:        "name": "EEA Czech Republic",
sources/de.json:        "name": "EEA Germany",
sources/dk.json:        "name": "EEA Denmark",
sources/es.json:        "name": "EEA Spain",
sources/fi.json:        "name": "EEA Finland",
sources/fr.json:        "name": "EEA France",
sources/gi.json:        "name": "EEA Gibraltar",
sources/hr.json:        "name": "EEA Croatia",
sources/hu.json:        "name": "EEA Hungary",
sources/ie.json:        "name": "EEA Ireland",
sources/mk.json:        "name": "EEA Macedonia",
```
* Tried some dry-runs, but nothing's happening?
```
node index.js -d -s "EEA Germany"
2017-08-15T08:44:40.714Z - info: --- Dry run for Testing, nothing is saved to the database. ---
2017-08-15T08:44:40.715Z - info: Running all fetch tasks.
2017-08-15T08:44:40.716Z - info: Dryrun completed, have a good day!
```
* You could use timezone names like `Etc/GMT+1` with `momentjs` for the `UTC-1`-like TZ conversion. See the [List of tz database time zones](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).

---

Somewhat related: an option to run a specific adapter would be useful. Like this:
```
node index.js --adapter eea-direct
```

maxgrossman:
@jflasher and @dolugen, see updates per your comments. 
I tested each eea source and times keep under 2 min. 
Once Battuta's nightly build is up and running I'll switch the stationsLink source

adapters/eea-direct.js
```diff
@@ -0,0 +1,238 @@
...
+  const coordinates = data[0];
+  const records = data[1];
+  parallel([
+    (done) => {
+      map(records, (record, cb) => {
+        let measurement = {
+          coordinates: makeCoordinates(coordinates, record[11]),
+          parameter: record[5].toLowerCase(),
+          date: makeDate(record[16], record[4]),
+          value: parseInt(record[19]),
```
jflasher: Why are we doing a `parseInt()` here? We're losing information by rounding to an int. Probably want to use `Number()`?

adapters/eea-direct.js
jflasher: Something still looks off with the times. Just ran this for DE and got the following for O3. Notice that every measurement has the same time.

```json
2017-08-21T17:21:25.986Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":26,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
2017-08-21T17:21:25.986Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":31,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
2017-08-21T17:21:25.986Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":35,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
2017-08-21T17:21:25.986Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":41,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
2017-08-21T17:21:25.986Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":56,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
2017-08-21T17:21:25.986Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":61,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
2017-08-21T17:21:25.986Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":63,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
2017-08-21T17:21:25.987Z - info: {"date":{"utc":"2017-08-19T23:00:00.000Z","local":"2017-08-20T00:00:00+01:00"},"parameter":"o3","location":"Halberstadt","value":66,"unit":"µg/m³","city":"Sachsen-Anhalt","attribution":[{"name":"EEA"}],"averagingPeriod":{"unit":"hours","value":1},"coordinates":{"latitude":51.896982,"longitude":11.057442},"country":"DE","sourceName":"EEA Germany","sourceType":"government","mobile":false}
```

I believe this means the data parsing is failing and you're just getting the midnight time. 

However, looking over this, we should be able to simplify. They're actually providing a time with a timezone, so that should have all we need I think? We can just parse `2017-08-20 12:00:00+01:00` as the local time and then also convert to UTC?

package.json
```diff
@@ -28,13 +28,15 @@
     "knex": "^0.13.0",
     "knex-postgis": "^0.2.2",
     "lodash": "^4.17.4",
+    "lodash.uniqby": "^4.7.0",
```
jflasher: Any reason to do this versus just use what's already in the lodash version already included?

package.json
```diff
@@ -28,13 +28,15 @@
...
     "moment": "^2.10.3",
     "moment-timezone": "^0.5.13",
     "pg": "^7.0.2",
     "proj4": "^2.3.14",
     "proj4js-defs": "0.0.1",
     "request": "^2.58.0",
     "require-dir": "^0.3.0",
+    "turf": "^3.0.14",
```
jflasher: Will need to break into separate modules.

adapters/eea-direct.js
olafveerman: All the sources report timestamps as `+01:00`, regardless of [where they are](https://github.com/openaq/openaq-fetch/issues/298#issuecomment-318382840). So we can't parse the date assuming that `+01:00` is the local time.

maxgrossman: good catch @olafveerman - since we have the lat/lon and moment like timezone strings, could we use something like [this](https://github.com/candu/efele-tz-world-geojson). We could do a spatial select (Getting polygon attrib) and use that as the tz string we pass to moment.

maxgrossman:
@jflasher @olafveerman I implemented a solution for dates that gets a timezone string using reverse geocoding via [tz-lookup](https://github.com/darkskyapp/tz-lookup), then converts dates, using the base time from the time field, to local and utc time respectively.

maxgrossman:
Per our new understanding that we can't rely on the fetched data's network timezones, the adapter now generates local timezones using [tz-lookup](https://github.com/darkskyapp/tz-lookup), a reverse geocoder for timezones. A note about tz-lookup. It generates the text string, for example 'Europe/Madrid'. So, we are relying on moment to handle any DST complications.

The actual implementation is as follows:

a `makeDate` func takes a record's time and parses along with its +01 utcoffset. the date obect from this parsing is set as the utc time.

For local, a string of the parsed date is passed along with thhe tz-lookup generated timezone to the `moment.tz()` method make the local date.

Below is a demonstration of these conversions for the 3 countries (portugal, poland, and latvia) @olafveerman discussed in #298

Portugal
```
from:
From: 2017-08-30 01:00:00+01:00
With local offset: +01:00
To:
Tue Aug 29 2017 20:00:00 GMT-0400 (EDT)
2017-08-30T01:00:00+01:00
```

Poland
```
From: 2017-08-29 07:00:00+01:00
With local offset: +02:00
To:
Tue Aug 29 2017 02:00:00 GMT-0400 (EDT)
2017-08-29T08:00:00+02:00
```

Latvia

```
From: 2017-08-29 11:00:00+01:00
With offset: +03:00
To:
Tue Aug 29 2017 06:00:00 GMT-0400 (EDT)
2017-08-29T13:00:00+03:00
```

olafveerman:
@jflasher I think all your comments are addressed, could you check and approve the changes?

@maxgrossman There is one last thing that I think we should address. The adapter parses the measurements into arrays and then relies on order to fetch values. By using `columns: true` on the parser, it returns objects, which allows us to do this:

``` js
parameter: record[pollutant].toLowerCase()
```
instead of: 
``` js
parameter: record[5].toLowerCase()
```

There are two benefits:

1. it makes the adapter more transparent and better to read for other people.
2. it makes the adapter more robust. If EEA decides to change the order of the columns, it may be hard to notice anything is wrong, or debug.

maxgrossman:
@jflasher - agree we should go for higher specificity. Stations current lack there of is probably due to how Battuta works. The most immediate fix I can think of is to change its [makeStation function](https://github.com/openaq/battuta/blob/master/utils/helpers.js#L148). 

Currently, this func (naively) just grabs the first of multiple responses from Pelias, not necessarily that with the most specific location. I can do some quick refactor to loop through each of the response objects and return the most specific. Also should clarify, specificity would be a property higher up in [this](https://github.com/openaq/battuta/blob/master/utils/helpers.js#L166) if statement chain.

I can also add in a backstop that first checks if the metadata record (which Battuta uses to generate records) includes a  `station_name`, the equivalent to the location here.

One thing to consider regarding picking the most specific of the geocoder results is that specificity does not necessarily mean accuracy. Each returned record has an attached ['confidence'](https://mapzen.com/search/explorer/?query=reverse&point.lat=48.858268&point.lon=2.294471&layers=locality) attribute. So, while going for specificity might get us to greater specificity, if we care about confidence, we might need to add in an internal confidence check (say greater than 0.6).

olafveerman:
Could we use the `station_name` that's passed as part of the measurement? The downside is that we can't guarantee friendly results. Andorra for example reports the following names for their three stations:

- Fixa
- Fixa oz
- Estacional oz Envalira

From a quick spot check, they do seem to be unique for a country.

Alternatively, we could use the station id as location name. It will be unique, and probably less subject to change over time.

@RocketD0g @jflasher Thoughts?

Sample record from Andorra:
``` json
{
  "network_countrycode": "AD",
  "network_localid": "NET-AD001A",
  "network_name": "",
  "network_namespace": "AD.GovernAndorra.AQ",
  "network_timezone": "http://dd.eionet.europa.eu/vocabulary/aq/timezone/UTC+01",
  "pollutant": "PM10",
  "samplingpoint_localid": "SPO-AD0942A-0005",
  "samplingpoint_namespace": "AD.GovernAndorra.AQ",
  "samplingpoint_x": "171336.05842057746",
  "samplingpoint_y": "5237637.567895099",
  "coordsys": "EPSG:4979",
  "station_code": "AD0942A",
  "station_localid": "STA-AD0942A",
  "station_name": "Fixa",
  "station_namespace": "AD.GovernAndorra.AQ",
  "value_datetime_begin": "2017-09-13 02:00:00+01:00",
  "value_datetime_end": "2017-09-13 03:00:00+01:00",
  "value_datetime_inserted": "2017-09-13 05:03:09+01:00",
  "value_datetime_updated": "+01:00",
  "value_numeric": "11",
  "value_validity": "1",
  "value_verification": "2",
  "station_altitude": "1080",
  "value_unit": "ug/m3"
}
```

olafveerman:
@jflasher `location` is still used to determine uniqueness of a measurement?

jflasher:
Combo of location, parameter, city and time I believe.

olafveerman:
@maxgrossman I don't think we should rely on Pelias given that uniqueness of a measurement is a 
> Combo of location, parameter, city and time

There is a chance that for two points in the same area Pelias will return the same information. We can use either:

1. `station_code`, which is less readable for a human, but likely more accurate and less prone to change over time
2. `station_name`, which may produce a more human readable result, though not guaranteed

My personal preference would be 1., but I'll defer to @jflasher and @RocketD0g who have a better sense of how important it is to have an understandable location name.

jflasher:
I don't have a strong preference here. I like seeing more friendly location names, but also feel that it's an ultimately flawed way to try and organize all the data. Think people should be more interested in lat/lon probably.

RocketD0g:
Agree with @jflasher.

maxgrossman:
1. sounds good to me.

olafveerman:
Just made that last change, I think it's ready to be pulled in @jflasher

jflasher:
:dancers: This looks good to go. Only question is, do we want to roll it out slowly? I think this turns it on for most all of the sources at once? @olafveerman or @maxgrossman if you want to leave only a few enabled and merge up to production, I'm good with that. Or we can roll the 🎲 and go for all of them 😁

jflasher:
Oh also, we really need to remember to squash and merge this one (but not on production)!

olafveerman:
I'd be fine doing a slow roll-out. How about deploying it this way:

1. Andorra - a new country with few measurements
2. Spain - an existing countries with a substantial amount of measurements
3. Rest of the existing sources
4. All of the new sources

@jflasher

jflasher:
Sounds great to me!

olafveerman:
This is happening!

Issue #182, opened 2016-07-30
Move from PostgreSQL to Elasticsearch?


jflasher:
There has been some talk of potentially moving from PostgreSQL to Elasticsearch. We're seeing a bit of a slowdown with Postgres, but I think that's more down to me not properly optimizing the database rather than anything inherent in the db. But as we accelerate number of datapoints we're collecting, is Postgres the final best answer? Should we use something like Elasticsearch? Or something else?

A couple of things to keep in mind as we think over other potential options
- We're collecting around 100K datapoints a day and that'll increase we we add a source like EU.
- We want to provide access to all historical datapoints
- There is also value in aggregation results (around all locations or cities or countries) so we'd need to be able to accommodate that
- We're going to be looking to bring on low-cost sensor data in the near future which could be a lot of data (open questions about how much we store, etc), but do we try to use the same system for those measurements or something different?

cc/ @dolugen @scisco @reinh

fpedrera:
In our team we have a couple of experiences dealing with tens of millions of rows in PostgreSQL/PostGIS under Carto (http://www.carto.com) and making spatial and aggregation queries. 

From our experience, we were able to optimize the performance by using these mechanisms: 

- Using composite indexes: https://www.postgresql.org/docs/9.5/static/indexes-multicolumn.html
- Using partial indexes: https://www.postgresql.org/docs/9.5/static/indexes-partial.html
- Working on the query itself, by using EXPLAIN ANALYZE on the query. 

Anyway, we are far for being experts, and we mainly followed the advice given by the Carto team and its CTO @javisantana. 

Anyway, some question arise: 
- What version of PostgreSQL are you using? Version 9.5 introduces some big data features: https://www.postgresql.org/about/news/1636/
- What kind of queries are you doing? We could take a look on them if you indicate the files on the GitHub repo with the queries. 

About moving to Elastic Search, I would first try to fine tune the PostgreSQL database before moving to manage an Elastic Search cluster.

jflasher:
Summarizing some other items that came up in Slack.

[...]

PR #354, opened 2017-07-31
Add feature to push measurements to S3

kamicut:
This commit allows the fetch process to push data measurements to S3
after adding them to the database.

The measurements are first added to the DB to determine if they're unique.
Unique records are then turned into CSV format and written to a user
specified AWS S3 bucket.

Pushing to S3 allows for:
1. Archiving the fetch data
2. Aggregation / Filtering in map-reduce pipelines using Amazon EMR or
Amazon Athena (could help with aggregations (#182))

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -11,7 +11,6 @@
 var argv = require('yargs')
   .usage('Usage: $0 --dryrun --source \'Beijing US Embassy\'')
   .boolean('dryrun')
-  .describe('dryrun', 'Run the fetch process but do not attempt to save to the database and instead print to console, useful for testing.')
   .alias('d', 'dryrun')
   .describe('source', 'Run the fetch process with only the defined source using source name.')
   .alias('s', 'source')
@@ -20,7 +19,7 @@ var argv = require('yargs')
   .alias('h', 'help')
   .argv;
 
-import { assign, filter, pick, chain, find } from 'lodash'; // eslint-disable-line import/first
+import { assign, filter, pick, chain, find, has } from 'lodash'; // eslint-disable-line import/first
 var async = require('async');
 var knex = require('knex');
 let knexConfig = require('./knexfile');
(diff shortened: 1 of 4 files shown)


kamicut:
@jflasher note: I haven't added the raw data measurement in the CSV output and have only added potential aggregation columns. This would discard attribution and averagingPeriod tags. Do you think we should add those as columns?

jflasher:
> I haven't added the raw data measurement in the CSV output and have only added potential aggregation columns. This would discard attribution and averagingPeriod tags. Do you think we should add those as columns?

I'd like S3 to become **the** authoritative source of the data (as opposed to the database). Because of this, somewhere in S3 we need to have all those fields. If the only downside is a bit more storage because we have all the data, let's just do that. 

We could also potentially keep the same format where we have major fields that we aggregate over and one JSON field. We will need to add fields in the future, so nice to have flexibility.

kamicut:
@jflasher, ok, i'll add that. I've had trouble adding the JSON as a column because the CSV parser in Athena gets confused with the commas and double quotes of a JSON object. I think we would be able to properly parse the row with a specific delimiter like `|` or a regex.

fetch.js
```diff
@@ -118,6 +117,43 @@ let buildSQLObject = function (m) {
...
+    latitude = m.coordinates.latitude || '';
+    longitude = m.coordinates.longitude || '';
+  }
+  row.push(latitude);
+  row.push(longitude);
+
+  // Add double quotes to each field
+  let quotedRow = row.map(item => `"${item}"`);
+
+  return quotedRow.join(',');
```
jflasher: Can we use something from http://csv.adaltas.com/ as we're already using `csv-parse` in the project? Think it'd be overall safer than making the CSV ourselves.

fetch.js
```diff
@@ -283,6 +326,59 @@ const saveFetches = function (timeStarted, timeEnded, itemsInserted, err, result
 };
 
 /**
+ * Create a key using the current datetime
+ * @param {Date} d javascript Date() object
+ * @return {string} string in format yyyy-mm-dd/unixtime
+ */
+const getDateKey = function (d) {
```
jflasher: We've already got a dependency on `moment` so feel free to just use `const now = moment();` if you want.

fetch.js
```diff
@@ -283,6 +326,59 @@ const saveFetches = function (timeStarted, timeEnded, itemsInserted, err, result
...
+    if (!validAWS) {
+      return done(new Error('missing AWS Credentials'));
+    }
+
+    log.info('Saving fetch to S3');
+
+    let s3 = new AWS.S3();
+
+    // Format into CSV
+    let rows = records.map(buildCSVObject).join('\n');
```
jflasher: Can maybe use http://csv.adaltas.com/ here as well?

kamicut:
@jflasher I'm thinking of having fetch just post raw JSON measurements and scrap the CSV part. We can move the CSV forming to `bumblebee`. What do you think?

jflasher:
Yeah, I think that'd be fine as long as the CSV was formatted to properly work with Athena (single line, etc). If we're transforming into something like Parquet for Athena, not an issue I guess.

kamicut:
Yes, we would have a JSON bucket for the raw data (and it could have any number of fields). We then have a derived bucket that has CSV data that we can use for Athena aggregations or other downstream APIs. I have read about Parquet for Athena, but I think we would need another EC2 creating the files, because it requires a bit more processing that a lambda function can handle (if I'm not mistaken).

jflasher:
> I have read about Parquet for Athena, but I think we would need another EC2 creating the files, because it requires a bit more processing that a lambda function can handle (if I'm not mistaken).

Yeah, I think this is true.

fetch.js
```diff
@@ -11,7 +11,6 @@
 var argv = require('yargs')
   .usage('Usage: $0 --dryrun --source \'Beijing US Embassy\'')
   .boolean('dryrun')
-  .describe('dryrun', 'Run the fetch process but do not attempt to save to the database and instead print to console, useful for testing.')
```
jflasher: Did this get removed on purpose?

kamicut: Ah, no, that's weird. Good catch!

jflasher:
This should be good to go, I'll test it out in the near future and look to get it merged in.

jflasher:
Looked this over locally and it looked good so going to merge to staging then should go up to master.

PR #356, opened 2017-08-01
Adapter/anhui

eliLitwack:
Given that pm25.in hasn't gotten back about the API key, wrote an adapter to scrape one of their sources directly.

This adds one dependency, transliteration, which romanizes non Latin characters - this is helpful for adding location names for stations (whose names are served in Chinese characters) without having a long table.

I also added a file to the data scripts directory which has the lat/lon for almost every measuring station in China - I'm not sure if that was the right directory for that.


jflasher:
Thanks! I think the big question for the community about this is whether we want to start pulling in individual sources or potentially wait for some other aggregated source we can use.

jflasher:
@RocketD0g any thoughts on this one?

RocketD0g:
If I understand right, this scrapes a government source directly. I think that's awesome, and let's go for it. I think it will be awhile before we can find a way to access the Chinese data in aggregate from a government source. Thanks for this contribution!

jflasher:
Thanks @eliLitwack! Added in https://github.com/openaq/openaq-fetch/pull/419

PR #358, opened 2017-08-09
Israel source support for Envista adapter

### LINKED ISSUE #146: Data from Israel

dolugen:
Resolves #146 

Added in conditionals for handling different sources. Richards Bay output OK as before. Israel output has varying numbers (13k - 20k) with random errors like before.

I've discussed with Max the idea of keeping a single Envista adapter and try to handle both sources, and we agreed to do it this way.

JF ninja edit: closes #328


jflasher:
If we can use the same adapter for the IL and ZA sources, I like that. Will the IL source get added to this PR as well? Can we close the other IL adapter PR?

dolugen:
Both sources are here. There are some things that should be moved from the other PR's branch, so don't yet close it.

maxgrossman:
I echo @jflasher - looks good. I did find coordinates for a handful of Israel stations that were missing them (via some googling).  Do we trust google and should I add/push new coords?

jflasher:
Just have a quick convo with @maxgrossman, the coords are for a city center, so we won't use them. It did come up that we're currently saving coords of 0,0 and we decided if a coord is 0,0, we will not save the location, but saving rest of the record is ok.

dolugen:
Removed null islands, and removed the older ZA adapter (thought I had renamed it).

dolugen:
Test run:

IL locations map

PR #359, opened 2017-08-09
Deploying UB data migration

jflasher:


jflasher:
Just a note here for my future reference (if I can find it!). The migration did not finish in a normal fetch-length of time, so it got canceled. To run the migration:

1. Disabled openaq-fetch 10min trigger in Lambda
2. Spun up a t2.large (prob didn't need something this big), installed openaq-fetch
3. Add `acquireConnectionTimeout: 600000` to knexfile.js (on ec2 instance)
4. Ran migration from ec2 instance with following command `PSQL_POOL_MAX=100 node --max-old-space-size=6096 node_modules/knex/bin/cli.js migrate:latest` 

Steps 3 and 4 are basically to increase connection timeout and pool size since we are trying to insert a large amount of records. The above ran on ~300,000 records and finished without issue (note to self, run inside tmux so no disconnects).

PR #360, opened 2017-08-10
Add feature to push measurements to S3 (#354)

jflasher:
* Add feature to push measurements to S3

This commit allows the fetch process to push data measurements to S3
after adding them to the database.

The measurements are first added to the DB to determine if they're unique.
Unique records are then turned into CSV format and written to a user
specified AWS S3 bucket.

Pushing to S3 allows for:
1. Archiving the fetch data
2. Aggregation / Filtering in map-reduce pipelines using Amazon EMR or
Amazon Athena (could help with aggregations (#182))

* Move from CSV to raw JSON

* Minor fixes

* Matching env var names and changing key name

### DIFF
--- a/.build_scripts/ecs-task-definition.json
+++ b/.build_scripts/ecs-task-definition.json
@@ -1,5 +1,6 @@
 {
     "family": "openaq-fetch",
+    "taskRoleArn": "arn:aws:iam::470049585876:role/uploadsS3",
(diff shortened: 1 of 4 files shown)


jflasher:
data is now going into http://openaq-fetches.s3.amazonaws.com/

cc/ @kamicut

PR #366, opened 2017-08-25
Adapter/campania

maxgrossman:
add adapter for campania.


jflasher:
Just seeing this now somehow, @maxgrossman is this ready to be looked at?

PR #370, opened 2017-09-12
Parse measurements as objects

olafveerman:
This addresses [the comment](https://github.com/openaq/openaq-fetch/pull/336#issuecomment-328865402) in the direct-eea PR and parses the measurements as objects instead of arrays.

@maxgrossman Can you double check that I'm using the right keys?


maxgrossman:
@olafveerman - the keys match what I'd been parsing.  going to merge in.

PR #372, opened 2017-09-15
Enable EEA Direct for Spain

olafveerman:
This enables EEA Direct for Spain and is the second step in the roll-out plan.

This PR can be merged once it's confirmed that Andorra has good results from EEA Direct.


olafveerman:
The Andorra data seems to be flowing in nicely:

image

Time to pull in Spain.

PR #373, opened 2017-09-15
Enable EEA Direct for all existing sources

olafveerman:
Part of the slow rollout of EEA Direct. Merge in after it's confirmed that results for Spain look good. (#372)


olafveerman:
Spain looks good apart from missing PM2.5 values. This is addressed in a separate PR: #383 

Pulling these adapters in.

PR #380, opened 2017-09-16
Hotfix: remove `results` subkey from task results

kamicut:
This fixes the underlying cause of openaq/openaq-api#351 which was
introduced by openaq/openaq-fetch#354

Each fetch task generated a message that contained information about
that task. To migrate to S3, I attached measurement results to these
messages. However I forgot to remove the property before pushing these
fetches to the PG database. This results in a bloated fetches table that
affects the API availibility for the fetch endpoint.

cc @olafveerman

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -20,7 +20,7 @@ var argv = require('yargs')
   .alias('h', 'help')
   .argv;
 
-import { assign, filter, pick, chain, find, has } from 'lodash'; // eslint-disable-line import/first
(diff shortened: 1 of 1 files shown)


jflasher:
Love that we need to remove the `results` key inside the `results` key... who names these things? :)

kamicut:
😅 yeah sorry about that, it did cause confusion on the API side because of the triply nested `results` keys as well.

PR #398, opened 2017-09-22
Up the EEA limit

olafveerman:
30 minutes is not enough. Files are generated once every hour. Filter the measurements to 1 hour + a small buffer.


jflasher:
I feel like the 30 minute cutoff would still be fine? fetch runs every 10 minutes, so there are around 3 fetches that'd run and be able to catch the new measurements every hour?

olafveerman:
The file itself is generated by EEA once an hour. A file generated/published at 13:00, can contain measurements that were inserted between 12:00 and 13:00.

jflasher:
Oh, got it!

PR #407, opened 2017-10-12
Arpae adapter

### LINKED ISSUE #302: Emilia-Romagna (Italy) data
near-real-time data:
https://dati.arpae.it/dataset/qualita-dell-aria-rete-di-monitoraggio/resource/a1c46cfe-46e5-44b4-9231-7d9260a38e68

to decode the parameters codes:
https://dati.arpae.it/dataset/qualita-dell-aria-rete-di-monitoraggio/resource/6fb8af6d-a3cc-4cf7-b098-ed95c5744cee?inner_span=True

stations (coordinates are EPSG:25832):
https://dati.arpae.it/dataset/qualita-dell-aria-rete-di-monitoraggio/resource/6b1c73c6-2fa9-40b2-854a-d3f822d1451a?inner_span=True

dolugen:
Resolves #302 

TODO:
- [x] Add data source
- [x] Prep locations metadata
- [x] Prep parameters metadata
- [x] Write the adapter itself
- [x] Fine tune time ranges in query
- [x] Set averagingPeriod
- [x] [Fix timezone](https://github.com/openaq/openaq-fetch/issues/302#issuecomment-337037647)

### DIFF
--- a/adapters/arpae.js
+++ b/adapters/arpae.js
@@ -4,53 +4,77 @@ import { REQUEST_TIMEOUT } from '../lib/constants';
 import { default as baseRequest } from 'request';
 import { default as moment } from 'moment-timezone';
 import { acceptableParameters, convertUnits } from '../lib/utils';
-import { difference, flattenDeep, zip } from 'lodash';
 const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
 
 exports.name = 'arpae';
 
-const ckanResourceID = "a1c46cfe-46e5-44b4-9231-7d9260a38e68";
+const ckanResourceID = 'a1c46cfe-46e5-44b4-9231-7d9260a38e68';
 // fixme: date range in sql
 const sql = `SELECT * from "${ckanResourceID}" WHERE reftime >= '2017-09-28T23:59:00' AND reftime <= '2017-09-29T00:00:00' ORDER BY reftime DESC`;
 const sqlUrl = `https://dati.arpae.it/api/action/datastore_search_sql?sql=${sql}`;
 
 exports.fetchData = function (source, cb) {
-    request(sqlUrl, (err, res, body) => {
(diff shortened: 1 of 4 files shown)


adapters/arpae.js
```diff
@@ -0,0 +1,292 @@
...
+exports.name = 'arpae';
+
+const ckanResourceID = 'a1c46cfe-46e5-44b4-9231-7d9260a38e68';
+const searchUrl = 'https://dati.arpae.it/api/action/datastore_search';
+
+exports.fetchData = function (source, cb) {
+  const yesterday = moment.tz(source.timezone).subtract(1, 'days').format('Y-M-D');
+  // not doing paging right now, so set a limit
+  const limit = 100 * acceptableParameters.length;
+  const queryUrl = `${searchUrl}?resource_id=${ckanResourceID}&filters={"reftime": "${yesterday}"}&limit=${limit}`;
```
jflasher: After some testing, it looks like this reftime gives the time AT the time passed, not the time SINCE. So basically when I put in 2017-10-10, I see only things at 2017-10-10T00:00:00 not since then. Do you see that too?

adapters/arpae.js
```diff
@@ -0,0 +1,292 @@
...
+        parameter: parameter,
+        date: {
+          utc: mDate.toDate(),
+          local: mDate.format()
+        },
+        country: source.country,
+        city: city,
+        location: location,
+        coordinates: coordinates,
+        attribution: [
```
jflasher: Can we get averagingPeriod added in here somewhere? Is it 1 hr?

dolugen: Oh right.

adapters/arpae.js
dolugen: You're absolutely right. I'll switch it back to SQL.

PR #409, opened 2017-10-14
Add current package-lock file

dolugen:
Following `npm install`'s advice. Generated from latest develop.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7330 @@
+{
+  "name": "openaq-fetch",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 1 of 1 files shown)


jflasher:
I might've actually ignore this in openaq-api. I guess we should just decide if we're going to use it or not. I've been against shrinkwrap before in other projects, but I believe this is supposed to be different, though I'm not sure how.

dolugen:
I think it'd be useful for a project like this. It sets an authoritative dependency tree for all users and CI tools.

It's similar to `npm-shrinkwrap.js`, but cannot be published. So if the project is a dependency of something, the lock file wouldn't be used.

Issue #418, opened 2017-11-02
Hong Kong AQ data!


RocketD0g:
Looks like station-level AQ data in physical units is available for Hong Kong here: 

http://www.aqhi.gov.hk/en/aqhi/past-24-hours-pollutant-concentration.html

Pollutants and units are defined, as well as include station names. 

Historical measurements are available through download here:
http://epic.epd.gov.hk/EPICDI/air/station/?lang=en

I'm not sure on coordinates, but may be able to be grabbed from the map?
http://epic.epd.gov.hk/EPICDI/air/station/?lang=en
Or from here: http://www.aqhi.gov.hk/en/monitoring-network/air-quality-monitoring-stations.html

Addresses of the stations also appear to be available here (Appendix B1):
http://www.aqhi.gov.hk/api_history/english/report/files/AQR2016e_final.pdf

If someone wants to build the adapter, we can create a list of the stations based on those station addresses.

dolugen:
It would be nice to have the coordinates ready. Any way to find out what they are?

RocketD0g:
Asking a colleague who mentioned this data/site.

RocketD0g:
@dolugen - No update on this, but I'm going to see if a volunteer would be up for creating this list of coordinates given the info above.

ymhuang0808:
According to the data portal in Hong Kong, the air quality data source is XML format. It would be better for parsing.

ymhuang0808:
I'm working on the HK adapter to fetch the [measurements](https://data.gov.hk/en-data/dataset/hk-epd-airteam-past24hr-pc-of-individual-air-quality-monitoring-stations) from HK open data. And I need a few time to implement it.

PR #432, opened 2017-12-20
Add an adapter for Hong Kong

ymhuang0808:
The adapter fetches data source from [here](http://www.aqhi.gov.hk/epd/ddata/html/out/24pc_Eng.xml), parses and converts it into OpenAQ format.

Hong Kong data source including the measurements within 24 hours. The adapter fetches the data according to `lastBuildDate` in [line 54](https://github.com/openaq/openaq-fetch/compare/develop...ymhuang0808:hk-adpater?expand=1#diff-23a2baf5cf8e024dbdbc2272bf1d08a0R54). 

Please take a look.

Related issue #418

### DIFF
--- a/adapters/hong-kong.js
+++ b/adapters/hong-kong.js
@@ -1,0 +1,158 @@
+/**
+ * This code is responsible for implementing all methods related to fetching
+ * and returning data for the Hong Kong data source.
+ */
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import cheerio from 'cheerio';
+import { default as moment } from 'moment-timezone';
+import { convertUnits } from '../lib/utils';
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+
+exports.name = 'hong-kong';
+
+/**
+ * Fetches the data for a given source and returns an appropriate object
+ * @param {object} source A valid source object
+ * @param {function} cb A callback of the form cb(err, data)
+ */
+exports.fetchData = (source, cb) => {
+  request(`${source.url}/24pc_Eng.xml`, function (err, res, body) {
+    if (err || res.statusCode !== 200) {}
+
+    try {
+      const data = formatData(body);
+
+      if (data.length === 0) {
+        return cb({message: 'Failure to parse data.'});
+      }
+      return cb(null, data);
+    } catch (e) {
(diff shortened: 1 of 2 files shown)


jflasher:
Thanks! I'll try to review next week if no one gets to look at it earlier.

ymhuang0808:
Sure! 😄

adapters/hong-kong.js
```diff
@@ -0,0 +1,158 @@
...
+const formatData = function (data) {
+  let measurements = [];
+  let $ = cheerio.load(data, { xmlMode: true });
+  const rootElement = $('AQHI24HrPollutantConcentration');
+  // Gets the last build date for getting current measurements
+  const lastBuildDate = moment.tz(rootElement.children('lastBuildDate').text(), 'ddd, D MMM YYYY HH:mm:ss Z', 'Asia/Hong_Kong');
+
+  // Traverses the AQHI24HrPollutantConcentration > PollutantConcentration and
+  // filters the measurements in lastBuildDate
+  rootElement.children('PollutantConcentration').filter(function (index, element) {
```
jflasher: I like the `filter` here to cut down on number of measurements that try to get inserted. However, `moment` seems to have trouble parsing the date and so the filter wasn't working correctly. I spent a fair amount of time trying to troubleshoot, but couldn't figure it out. Could you change the logic so that the `DateTime` for each measurement is just checked against the current time (should probably use UTC for this check) and sees that it's less than 2 hours old?

ymhuang0808: @jflasher Sorry for late reply. Since, we may not fetch duplicated measurements, I think the logic is filtered by `lastBuildDate` is better.

ymhuang0808: I dont't find the parsing error in the filter when I'm performing a dry run. Could you paste the error message?

jflasher: No worries about the late reply, thanks for working on the adapter! Just to note, the adapters do not need to ensure uniqueness, the database will do that. So the adapter can grab duplicates and pass them along, but it's better if it doesn't try to insert alot of measurements that will likely be duplicates, and it adds database overheard to check for uniqueness and slows down inserts.

For the date piece, nevermind what I said! I wrote this nice long explanation and then said maybe it was just a version issue with `moment`. I upped the version and it works fine! Will change the package.json and merge in from a different branch. Thanks!

PR #520, opened 2018-09-05
Streamed fetch

MichalCz:
Here's a major rewrite of openaq-fetch.

### Major changes:

- simplification of the main `fetch.js` file by extracting all actual methods to library folder.
- separation of utility and fetch functions to separate files.
- introduction of streaming for database save and s3 streaming process
- more flexible adapters - measurements can be provided as any `streamable` object if exposing `fetchStream` method
- use of async/await in the fetch process
- compatibility with node 8
- additional documentation of formats in docs/
- additional command line options and switches

The `fetchStream` for an adapter is now much more flexible and can return (including asynchronously):

- `Array<Measurement>` - the same data that `fetchData` would return.
- `Iterable<Measurement>` - any iterable
- `AsyncGeneratorFunction<Measurement>` - in essence, `async function*(source) { ... }`, an [async generator](https://github.com/tc39/proposal-async-iteration#async-generator-functions)
- `Readable<Measutement>` - a readable stream that will be streamed to all endpoints.

### Benefits

The change introduces streaming throughout all the data fetching process from acquisition (on selected adapters) to database save and s3 streaming.

Here's the last output from a full fetch on an Intel i7 5820k, 16GB RAM machine:

        Command being timed: "node . -v"
        User time (seconds): 82.43
        System time (seconds): 4.96
        Percent of CPU this job got: 56%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:34.48
        Maximum resident set size (kbytes): 302696
        Minor (reclaiming a frame) page faults: 134670
        Voluntary context switches: 120780
        Involuntary context switches: 1538
        File system outputs: 8832
        Page size (bytes): 4096

Improvements include: 

- overall execution time - the measurements are saved to DB and streamed to s3 during fetch process.
- CPU usage - the transformations no longer operate on large arrays, there's also less garbage collection needed.
- memory usage - if the adapter fully supports streaming, the data is never aggregated and kept in memory as whole.


jflasher:
Thanks @MichalCz, I'll be checking this out over the next few days!

MichalCz:
Ok. In mean time I did find some issues with Error handling, that is causing the program to exit despite not having set the `strict` flag nor env...

lib/measurement.js
```diff
@@ -0,0 +1,227 @@
...
+      ;
+    }
+  } catch (e) {
+    out.end(new FetchError(ADAPTER_ERROR, source, e, 0));
+  }
+
+  return createFetchObject(out, source, dryrun);
+}
+
+export function printDataToLog (stream, log) {
```
jflasher: I noticed when running `node index.js -d -s 'Beijing US Embassy' -v` I'm seeing output like 

```
2018-09-10T12:45:18.742Z - verbose:  utc=Mon Sep 10 2018 08:00:00 GMT-0400 (EDT), local=2018-09-10T20:00:00+08:00, parameter=pm25, value=41, unit=µg/m³, value=1, unit=hours, location=Beijing US Embassy, city=Beijing, country=CN, latitude=39.95, longitude=116.47, attribution=[name=StateAir.net, url=http://www.stateair.net/web/post/1/1.html], sourceName=Beijing US Embassy, sourceType=government, mobile=false
```
which is not stringified. For this output, I prefer stringified as it'll convert the UTC time properly which makes it easier to examine (instead of it taking into account your local time zone when it's an object). The output at this code block is being stringified, but this must not be where this output is happening for the verbose logging.

MichalCz:
Noted, working on all of those. :)

MichalCz:
Ok, so one thing to add to the list above - there's an issue with how Babel works out if an object is `instanceof` something derived from `Error` - the issue is that it doesn't work at all.

So now sadly any Error is simply thrown and I need to make a small workaround. After that I'll get to the list above.

lib/measurement.js
MichalCz: Hmm... the "why" is quite simple - the schema says that one should be a Date object, second a String:

```
        "date": {
            "type": "object",
            "additionalProperties": false,
            "required": ["utc", "local"],
            "properties": {
                "utc": {
                    "type": "date"
                },
                "local": {
                    "type": "string",
                    "pattern": "\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(\\+|-)\\d{2}:\\d{2}"
                }
            }
        },
```

So when that's stringified you get different valuies.

I'll check if that doesn't pose any additional issues, but could we simply use two strings here? The database should be ok with that and the strings will be the same - and I'll add mapping for old format.

docs/source.md
```diff
@@ -0,0 +1,39 @@
...
+### Source object properties
+
+The source object must consist of the following properties:
+
+* `url (String)` - the *informative* url about the source,
+* `adapter (String)` - the module name of the adapter a file or directory residing in `./adapters`,
+* `name (String)` - an *informative* source name
+* `country (String)` - the country for which the data is provided
+* `description (String)` - description of the source
+* `contacts (Array<String>)` - a list of e-mail addresses for contact about the source
```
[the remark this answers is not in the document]
MichalCz: I can set a default value of that if you like, but if you'd want to use that field in future - I personally would leave it as is.

docs/source.md
```diff
@@ -0,0 +1,39 @@
...
+* `active (Boolean)` - is the source active or not (if not it will be omitted in fetch process)
+
+Optional properties (used to add information to measurements):
+
+* `type (Enum<government|private|?>)` - Region name
```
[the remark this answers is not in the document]
MichalCz: Thanks. Will fix this. :)

docs/source.md
```diff
@@ -0,0 +1,39 @@
...
+* `city (String)` - City name
+* `location (String)` - Name of the location
+* `mobile (Boolean)` - Is the source mobile
+
+Additonally the whole object is passed to the adapter so anything can be added to the structure. Adapter usually use:
+
+* `sourceURL (String)` - the url to use to fetch the data from
```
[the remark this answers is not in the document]
MichalCz: Ok. Will fix.

fetch.js
```diff
@@ -7,451 +7,80 @@
...
+      // mark sources as started
+      .do(markSourceAs('started', runningSources))
+      // get measurements object from given source
+      .map(async (source) => getMeasurementsFromSource(source, env))
+      // perform streamed save to DB and S3 on each source.
+      .do(streamMeasurementsToDBAndStorage(doSaveToS3, env, bucketName))
+      // mark sources as finished
+      .do(markSourceAs('finished', runningSources))
+      // handle adapter errors to be forwarded to main stream and well handled.
+      .use(forwardErrors)
```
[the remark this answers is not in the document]
MichalCz: First of all - the method here is supposed to selectively forward errors from the measurements stream flowing from adapter to S3, database and console, to the stream that runs all adapters. Any error forwarded there gets handled by the `handleFetchErrors` method at the end of the file.

I'll document this a bit, but in general all errors coming from an adapter are handled as stream errors, that is:
 * an adapter fetch error (thrown as the FetchError class).
 * an asynchronous error.
 * wrong data format returned (not an array, data stream, generator etc.)
 * adapter not found.
 * measurement error.

In effect it should:
 - ignore adapter fetch errors (so simply mute them here and, after a small change I need to commit still, also log a warning)
 - ignore measurement errors and only log them.

As previously it won't ignore:
 - adapter runtime errors
 - adapter syntax errors
these are handled, but forwarded to the main stream.

This only changes if you turn on `--strict` mode, where any error should be forwarded to the main stream.

fetch.js
```diff
@@ -7,451 +7,80 @@
...
+        // TODO: filter out sources not included in fetch
+        reportAndRecordFetch(fetchReport, sources, env, apiURL, webhookKey)
+      );
+  })()
+])
+  .catch(
+    handleFetchErrors(log, env)
+  )
+  .then(
+    exitCode => process.exit(exitCode || 0)
```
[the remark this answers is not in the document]
MichalCz: Actually no, `handleFetchErrors` extracts the exitCode from the `FetchError` object - if there's something there it will be that (for instance `100`), but if that's not given a default of `0` is used.

fetch.js
```diff
@@ -7,451 +7,80 @@
...
+      .do(markSourceAs('finished', runningSources))
+      // handle adapter errors to be forwarded to main stream and well handled.
+      .use(forwardErrors)
+      // convert to measurement report format for storage
+      .map(prepareCompleteResultsMessage(fetchReport, runningSources))
+      // aggregate to Array
+      .toArray()
+      // save fetch log to DB and send a webhook if necessary.
+      .then(
+        // TODO: filter out sources not included in fetch
```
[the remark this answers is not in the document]
MichalCz: Yes, the ones that were made `filtered` - I forgot to ask you about this.

So when we run a single source - what should we save to the db?
And when we run only active sources - should we save only those?

So I guess we should remove them, but I had some second thoughts back then.

lib/utils.js
```diff
@@ -156,3 +68,65 @@ export function safeParse (json) {
...
+    return new Promise(() => 0); // never resolve
+  }
+}
+
+export function handleFetchErrors (log) {
+  return (error) => {
+    const cause = error instanceof FetchError ? error : error.cause;
+
+    if (cause instanceof FetchError) {
+      if (cause.is(STREAM_END)) return cause.exitCode || 0;
```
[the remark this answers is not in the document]
MichalCz: I hear you. :)

I think I actually don't normally send stacks to `Winston` now, but that's worth double checking. Only place where that happens is in the final error handler, before process.exit.

In any case I'll double check and report back.

lib/utils.js
```diff
@@ -1,134 +1,46 @@
...
-
-  for (var i = 0; i < measurements.length; i++) {
-    var m = measurements[i];
-
-    // Skip if we don't have a valid measurement
-    if (!m) {
-      continue;
-    }
+export function convertUnits (array) {
+  array.filter(x => x).forEach(unifyMeasurementUnits);
```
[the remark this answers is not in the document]
MichalCz: `x => x` in `filter` is simply to filter out falsy values, especially nulls or undefined. Those units cannot be unified, since there's nothing there to unify. ;)

BTW, I've taken the liberty of removing this method in a next commit (I left a stub, so that legacy adapters won't break). Instead it's being run by the fetch process, along with a couple other checks (like removing unwanted measurements).

lib/env.js
```diff
@@ -0,0 +1,112 @@
+
+let _argv = require('yargs')
+  .usage('Usage: $0 -d -s \'Beijing US Embassy\'')
+  .options('quiet', {
+    boolean: true,
+    describe: 'Show lots additional logging information (more than verbose)',
```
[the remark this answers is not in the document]
MichalCz: Very much so... ;)

MichalCz:
Joe, I need to push the dependencies update in the same pull request otherwise we'd start a major conflict and that would get worse and worse. So here's another group of commits.

fetch.js
```diff
@@ -7,451 +7,80 @@
  */
```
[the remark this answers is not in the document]
MichalCz: Hmm... Strangely I can't see 2048, but 2368 in my code. Let me push a couple new commits and let's see...

.build_scripts/bench-adapters.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
```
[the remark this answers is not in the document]
MichalCz: Actually I added that in the .es6 module...

MichalCz:
I need to rewrite the measurement test a little so it works again. I'll finish that tomorrow. In meanwhile if you could check my answers that'd be good. :)

docs/source.md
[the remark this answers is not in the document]
jflasher: Yeah, I think it's fine as is. When I originally read this, the description made me think it was contacts at the source of the data. But it looks fine to me.

fetch.js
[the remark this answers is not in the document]
jflasher: I think whatever gets run should get saved. I only ever run a single source for testing purposes, but would want just that source saved to the db. And if I'm only running active sources, would only want those saved.

fetch.js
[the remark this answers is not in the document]
jflasher: I'm now seeing 2688 for new code, 2899 for old code. Wonder if this is due to stricter rejection code somewhere?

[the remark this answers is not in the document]
MichalCz: That is possible - I did fiddle with the quality checks quite a lot, so it may be I missed some kind of lowercase. I'll check that.

lib/measurement.js
MichalCz: We would, but I simply added a date correction step along the way - so no need to change any adapter code.

In general I'm trying to make no changes at all to __all adapters__. Whatever worked before should work after.

MichalCz:
Joe - I've added ppb as measurement unit - was that intentionally left out as a difference between long billion (thousand millards) and short billion (thousand millions)?

fetch.js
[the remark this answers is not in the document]
MichalCz: It seems there has been some issues with `.batch(64)` in the adapter itself - it actually is a regression in `scramjet`.

I removed the flatmap as it's not needed anymore so I'll have more time looking for the batch issue.

fetch.js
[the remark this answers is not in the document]
MichalCz: Added.

lib/utils.js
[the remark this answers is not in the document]
MichalCz: No stacks are going to Winston except for when fetch.js actually crashes.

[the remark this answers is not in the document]
MichalCz: And that's a single stack, so it should be ok to leave it.

jflasher:
> @jflasher - I've added ppb as measurement unit - was that intentionally left out as a difference between long billion (thousand millards) and short billion (thousand millions)?

How do you mean you added it? At the end of the day, we should only end up with measurements using units of either `µg/m³` or `ppm`. If something comes in like `ug/m3`, `PPM`, we can do a simple name change. If it's like `g/cm3` or `ppb` or variants, we will do some math to convert to our preferred units. We don't convert across `µg/m³` and `ppm` and if something has a unit we don't understand, we just ignore it (or at least we should). The renaming and math were done in https://github.com/openaq/openaq-fetch/blob/develop/lib/utils.js#L92

MichalCz:
@jflasher I mean I added the ppb to ppm conversion. But if I read you correctly I will revert that change, please confirm.

jflasher:
I think this is probably fine as long as it’s done at some point before the final validation occurrs so we don’t lose any measurements because of it. And if you could keep the same conversion math as was in the previous method, that’d be great. While the math isn’t hard, it care out of looking at a bunch of the ways different sites did their labeling which is often a mess!

MichalCz:
@jflasher done. :)

MichalCz:
@jflasher please hold off the pull for another day or two. I've noticed as problem in error handling in scramjet and I'd need to fix it. Currently the errors are either completely muted or crash the fetch process.

Because of a small mistake, scramjet doesn't pass the error wrapped in a transport object when it occurs in another error handler. Instead it simply passes the actual error which is inconsistent, I did however rely on this to work. I'll release the new version of scramjet tomorrow and update this branch.

MichalCz:
Hello @jflasher,

All the fixes are now in. I've added some docs and removed some unused code left over from the refactor.

The adapters now nicely report errors. This is also now covered by automated tests.

lib/db.js
```diff
@@ -0,0 +1,173 @@
...
+  // Run any needed migrations and away we go
+  return _pg;
+}
+
+/**
+ * Build an object that can be inserted into our database.
+ * @param {object} m measurement object
+ * @return {object} an object capable of being saved into the PostgreSQL database
+ */
+function convertMeasurementToSQLObject (stream, pg) {
```
jflasher: Hi @MichalCz, looking over this this morning, and saw the following behavior when trying to run using the local database.

```
$ node index.js -s 'StateAir_Hanoi'                                                               [±:signicode-feature/streamed-libs:∂]
2018-09-24T11:58:49.974Z - info: --- Full fetch started. ---
2018-09-24T11:58:49.988Z - info: Looking up source StateAir_Hanoi
2018-09-24T11:58:50.547Z - info: Getting stream for "StateAir_Hanoi" from "stateair"
2018-09-24T11:58:51.018Z - info: Using connection data from knexfile.local
2018-09-24T11:58:51.112Z - info: Connecting to database.
2018-09-24T11:58:51.338Z - info: Database connected and migrations are handled, ready to roll!
2018-09-24T11:58:51.374Z - error:  message=insert into "measurements" ("attribution", "averagingPeriod", "city", "coordinates", "country", "date", "location", "mobile", "parameter", "sourceName", "sourceType", "unit", "value") values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13) returning "location" - column "attribution" of relation "measurements" does not exist, stack=error: column "attribution" of relation "measurements" does not exist
```

This is odd because https://github.com/openaq/openaq-fetch/pull/520/files#diff-be228df81b99a384aa50ad50bde7fca0R38 looks to be doing what I expect it should be doing, but the measurement trying to be inserted doesn't seem to be hitting it? Note how it doesn't include `date_utc` or `data` but does include `attribution`.

docs/dev-quick-start.md
```diff
@@ -0,0 +1,35 @@
...
+4. Install dependencies `cd openaq-fetch && npm install`
+5. Create a user and db in postgres in `docker exec -ti postgis psql -U postgres`
+```sql
+CREATE ROLE openaq WITH LOGIN PASSWORD 'openaq-password';
+CREATE DATABASE openaq OWNER openaq;
+\connect openaq
+CREATE EXTENSION postgis;
+\quit
+```
+6. Create a local knexfile `cp knexfile.js knexfile.local.js && editor knexfile.local.js`
```
jflasher: Change to `knexfile-local.js`

docs/index.md
```diff
@@ -0,0 +1,15 @@
+# Documentation
+
```
jflasher: Can you add a link to this somewhere in the README file? This is incredibly helpful and I don't want people to miss it.

MichalCz: Doing.

docs/dev-quick-start.md
MichalCz: Doing

lib/db.js
MichalCz: I'll check that... this is the first column so it may be that the whole table is somehow badly created... Maybe my knexfile-local.json causes that? Or db is not right? I'll get back to you on this.

jflasher:
I was using my existing db and the migrations didn’t change so I’m thinking it has something to do with the way the object to be inserted is created rather than anything on the db side?

MichalCz:
You may be right. I recall there was a wrapping function somewhere there. I'll check it.

lib/db.js
MichalCz: Fixed.

docs/index.md
MichalCz: Done.

docs/dev-quick-start.md
MichalCz: Done.

MichalCz:
@jflasher I also added a small change to result messages pushed out to logs (including number of duplicates and time elapsed).

lib/db.js
```diff
@@ -0,0 +1,182 @@
...
```
jflasher: Trying the latest, when running with command below, I'm getting an error.

```
$ node index.js -s 'StateAir_Hanoi' -d                                                            [±:signicode-feature/streamed-libs:∂]
2018-09-26T13:41:29.746Z - info: --- Dry run for Testing, nothing is saved to the database. ---
2018-09-26T13:41:29.775Z - info: Looking up source StateAir_Hanoi
2018-09-26T13:41:30.276Z - info: Getting stream for "StateAir_Hanoi" from "stateair"
2018-09-26T13:41:30.685Z - info: Using connection data from knexfile-local
2018-09-26T13:41:30.756Z - info: Connecting to database.
2018-09-26T13:41:30.958Z - info: Database connected and migrations are handled, ready to roll!
2018-09-26T13:41:31.164Z - error: Runtime error occurred in DataStream(unused): Error: Cannot read property 'utc' of undefined
    at Promise.all.catch (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/util/mk-transform.js:59:44)
    at <anonymous>
  caused by:
    at DataStream.stream.map.data (/Users/jflasher/workspace/openaq-fetch/lib/db.js:50:17)
    at <anonymous>
  --- raised in DataStream(unused) constructed ---
    at PromiseTransformStream (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/util/promise-transform-stream.js:65:27)
    at DataStream (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/data-stream.js:55:9)
    at DataStream.map (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/data-stream.js:153:26)
    at convertMeasurementToSQLObject (/Users/jflasher/workspace/openaq-fetch/lib/db.js:41:17)
    at DataStream.use (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/data-stream.js:304:20)
    at streamRecordsToPg (/Users/jflasher/workspace/openaq-fetch/lib/db.js:75:6)
    at DataStream.use (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/data-stream.js:304:20)
    at streamDataToDB (/Users/jflasher/workspace/openaq-fetch/lib/db.js:93:6)
    at /Users/jflasher/workspace/openaq-fetch/lib/db.js:178:9
    at <anonymous>
```

I put a log statement to see what the `data` variable was getting. In the case where it errors, it's getting the below which looks like the already converted object. So it looks like this function may be getting called multiple times and will fail if it runs on already converted data, which I suppose is good! 

Note also that the coordinates on the converted object is the full postgis coordinates object, but just needs to be a simple lat/lon object.

```
{ location: 'US Diplomatic Post: Hanoi',
  value: 17,
  unit: 'µg/m³',
  parameter: 'pm25',
  country: 'VN',
  city: 'Hanoi',
  source_name: 'StateAir_Hanoi',
  date_utc: '2018-09-26T12:00:00.000Z',
  source_type: 'government',
  mobile: false,
  data: 
   { date: 
      { utc: '2018-09-26T12:00:00.000Z',
        local: '2018-09-26T19:00:00+07:00' },
     parameter: 'pm25',
     value: 17,
     unit: 'µg/m³',
     averagingPeriod: { value: 1, unit: 'hours' },
     location: 'US Diplomatic Post: Hanoi',
     city: 'Hanoi',
     country: 'VN',
     coordinates: { latitude: 21.021938, longitude: 105.81881 },
     attribution: [ [Object] ],
     sourceName: 'StateAir_Hanoi',
     sourceType: 'government',
     mobile: false },
  coordinates: 
   Raw {
     client: 
      Client_PG {
        config: [Object],
        logger: [Object],
        connectionSettings: [Object],
        driver: [Object],
        pool: [Object],
        valueForUndefined: [Object],
        _events: [Object],
        _eventsCount: 4,
        makeKnex: [Function: makeKnex],
        version: '9.5.3' },
     sql: 'ST_geomFromText(?, ?)',
     bindings: [ 'Point(105.81881 21.021938)', 4326 ],
     _wrappedBefore: undefined,
     _wrappedAfter: undefined,
     _debug: undefined } }
```

MichalCz: I'll check it today or tomorrow evening (should be fixed tomorrow your time).

jflasher:
Thanks!

lib/db.js
MichalCz: Oh, now I see my misunderstaning. It's the `-d` - a small fix resolved this.

MichalCz: So this was due to the fact that I removed an unneeded argument from one of the methods and didn't change the call. Now this will work as it should.

MichalCz:
@jflasher I added test for the following command: `node . -ds test_adapter` to the tests run. This way a mistake like the one fixed yesterday should not go past the tests. This way there will also be a simple confirmation if the fetch process still works correctly after any new changes.

lib/db.js
jflasher: Just dropped the `-d` and it's still happening, same error.

MichalCz: I found it now. It didn't occur if the data was not streamed to S3. As you suspected the object was transformed to the SQL version twice.

jflasher: Hi @MichalCz, I wanted to make sure I should be waiting for another commit before checking fetch behavior? I didn't see any commit since your last comment and wanted to make sure I didn't miss something. Thanks!

MichalCz: Hi @jflasher I didn't push to the right remote after I fixed the issue. :(

There's that commit now.

jflasher: Great, this looks to be working now, thanks!

lib/db.js
```diff
@@ -0,0 +1,181 @@
...
+    if (dryrun) {
+      return stream
+        .do(m => log.verbose(JSON.stringify(m)))
+        .run();
+    } else {
+      const key = `realtime/${moment().format('YYYY-MM-DD/X')}.ndjson`;
+      const pg = await getDB();
+
+      await Promise.all([
+        doSaveToS3 ? saveResultsToS3(stream, new (require('aws-sdk').S3)(), bucketName, key) : true,
```
jflasher: Was looking to test the data format of data saved to S3. Ran with the following command (using valid AWS key/secret)

`AWS_ACCESS_KEY_ID=ID AWS_SECRET_ACCESS_KEY=KEY AWS_BUCKET_NAME=openaq-fetches SAVE_TO_S3=true node index.js` 

and was seeing a lot of the following errors

```
(node:29353) UnhandledPromiseRejectionWarning: Error: Cannot read property 'Symbol(Symbol.iterator)' of undefined
    at promise.catch (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/util/promise-transform-stream.js:221:31)
    at process._tickCallback (internal/process/next_tick.js:178:7)
  caused by:
    at into (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet/lib/data-stream.js:380:35)
    at process._tickCallback (internal/process/next_tick.js:178:7)
  --- raised in DataStream(unused) constructed ---
    at new PromiseTransformStream (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/util/promise-transform-stream.js:65:27)
    at new DataStream (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet-core/lib/data-stream.js:55:9)
    at DataStream.flatMap (/Users/jflasher/workspace/openaq-fetch/node_modules/scramjet/lib/data-stream.js:373:21)
    at saveResultsToS3 (/Users/jflasher/workspace/openaq-fetch/lib/db.js:108:6)
    at /Users/jflasher/workspace/openaq-fetch/lib/db.js:176:22
    at process._tickCallback (internal/process/next_tick.js:178:7)
(node:29353) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 186533)
```

Tried again without the extra `SAVE_TO_DB` and `AWS_BUCKET_NAME` vars and everything worked fine again, so I presume it's tied to the output there. What's interesting is that this didn't happen at the end of the fetch which is when I presume the data is getting written to S3, but happened in mid-fetch as the adapters were running.

lib/db.js
```diff
@@ -0,0 +1,181 @@
...
+ */
+export async function saveFetches ({timeStarted, timeEnded, itemsInserted, errors, results}) {
+  const pg = await getDB();
+
+  return pg('fetches')
+    .insert([{
+      time_started: new Date(timeStarted),
+      time_ended: new Date(timeEnded),
+      count: itemsInserted,
+      results: JSON.stringify(errors || results)
```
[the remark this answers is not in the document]
MichalCz: I guess I wanted to get too much info there. Will look into this tomorrow.

lib/db.js
MichalCz: The happening in middle part is expected - steaming to s3 happens white adapters were still running. I'll check this out tomorrow.

jflasher:
Ah right, forgot that we were streaming here!

lib/db.js
MichalCz: Ok. For some reason I didn't merge some of the work from the s3 branch and that was an old version which didn't have a slight chance to work.

@jflasher - there's a limitation of s3 that requires the uploads to have file sizes set before the upload - trouble is, we don't actually know how big the file will be. Therefore the stream is chunked. In the upload code I set a chunk to 1MB - which sets a maximum of 10GB per file (S3 allows a maximum of 10k chunks).

Now I thought this should be controller via env, so let me add a quick commit for that.

MichalCz: Ok, that's also pushed.

lib/db.js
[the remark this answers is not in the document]
MichalCz: Hi Joe,

Indeed the alternative `(errors || results)` was out of place (especially that the errors are included the results object). This has been fixed, postgre now shows:

```sql
select * from fetches;

-[ RECORD 1 ]+--------------------------------------------------------------------------------------------------------------------------------------
id           | 49
time_started | 2018-10-04 14:14:48.997+00
time_ended   | 2018-10-04 14:14:51.324+00
count        | 24
results      | [{"message":"New measurements found for StateAir_Hanoi: 24","failures":{},"count":24,"duration":0.266,"sourceName":"StateAir_Hanoi"}]
```

lib/db.js
```diff
@@ -162,8 +177,10 @@ export async function saveSources (sources) {
  * @param {OpenAQEnv} env
  * @param {String} bucketName
  */
-export function streamMeasurementsToDBAndStorage ({doSaveToS3, dryrun}, bucketName) {
-  return async ({ stream, counts }) => {
+export function streamMeasurementsToDBAndStorage (stream, {doSaveToS3, doSaveToDB, dryrun}, bucketName) {
```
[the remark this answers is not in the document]
MichalCz: Now all the data is saved indeed. I'll chain s3 save after db save and add a filter for duplicates. ETA today.

lib/env.js
```diff
@@ -96,6 +96,7 @@ export const getEnv = () => {
   const processTimeout = _env.PROCESS_TIMEOUT || 10 * 60 * 1000; // Kill the process after a certain time in case it hangs
   const bucketName = _env.AWS_BUCKET_NAME || '';
   const doSaveToS3 = _env.SAVE_TO_S3 === 'true' || +_env.SAVE_TO_S3;
+  const doSaveToDB = !('SAVE_TO_DB' in _env) || _env.SAVE_TO_DB === 'true' || +_env.SAVE_TO_DB;
```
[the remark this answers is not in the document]
MichalCz: I used that for testing only. But your other comment clears this out. I'll remove this.

lib/db.js
MichalCz: Yeah. I did note that. So I'm rewriting the save to s3 so that all the measurements are sent to a single stream constructed before the fetch is started - and that stream will get saved.

I got this working, but at the same time I'll need to spend little more time to make it look better.

lib/db.js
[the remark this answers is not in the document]
MichalCz: Fixed in the last commit. Only saved items are being pushed to s3.

lib/env.js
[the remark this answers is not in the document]
MichalCz: Removed.

MichalCz:
@jflasher as promised here's the updated version of s3 save with:

* Only one s3 file per fetch with all data
* S3 file is now closed at the end, which required me to add a new, rather helpful thing: `errors.js:cleanup.add` - calling this method will allow you to run any asynchronous function before the process exits (for example: make sure that s3 is saved before we call process.exit()). May come in handy in other cases.
* Added small improvements in log messages in debug mode.
* Removed `doSaveToDB` toggle

lib/db.js
```diff
@@ -110,11 +111,25 @@ function streamDataToDB (stream, pg, counts) {
...
-  stream
+
+  const finishing = stream
+    .catch(ignore)
+    .map(({status, ...measurement}) => measurement)
     .use(streamDataToS3, s3, bucketName, key, s3ChunkSize)
     .catch(e => output.raise(e));
 
-  output.whenEnd().then(() => stream.end()).catch(ignore);
+  const endStream = () => {
```
jflasher: I'm still seeing a timeout when running the following command

`AWS_ACCESS_KEY_ID=ID AWS_SECRET_ACCESS_KEY=KEY AWS_BUCKET_NAME=openaq-fetches SAVE_TO_S3=true node index.js`

there are no running adapters and looking at the db, there are no new measurements being inserted for a long time before timeout, so I wonder if this is still not getting called?

MichalCz: Not this actually, but I have a hunch...

MichalCz: Hi Joe - I found the issue in an implementation of `empty` method in scramjet. I should fix it by tomorrow morning your time.

MichalCz: Hi again. Just to let you know, the issue I found seems more complex to solve because of conflicting requirements in couple different methods - so I'm working on this currently, but it's taking more than than anticipated.

MichalCz: Hi Joe, the commit I just pushed should fix the issue.

MichalCz:
Hi @jflasher,

How's the check going? Did you manage to run this version on staging?

Best,
M.

jflasher:
Sorry @MichalCz was disconnected for the last few days, will look at this in the next couple of days.

jflasher:
Saving to database and S3 looks fine, so I'm going to merge this into staging and see how it runs there.

PR #524, opened 2018-10-13
Adapter: China AQI Data

MichalCz:
The adapter work is straight from the adapters/china branch, this here is just a streamed rewrite of that branch.

The adapter itself is very simple, just streams the data in intermediate format from the s3 bucket (lines 33-40) then converts the measurements structure, adapts units (lines 41, 42) and flattens the resulting arrays to a stream of single measurements.


MichalCz:
@jflasher checked and ready to merge.

PR #525, opened 2018-10-13
Adapter: arpalazio streamed

MichalCz:
This is the streamed rewrite of the `arpalazio` adapter.

Te adapter works based on two steps:

 * obtaining the links to data urls from the website using `cheerio` - this was largely untouched, just rewritten to async/await.
 * fetching and parsing the fixed width columns - this has been fully rewritten to streamed version.

List of sites has been moved out of adapter code to `"metadata"` key in `sources/it.json`.

Suggestions for later:

 * current handling of year may lead to having to manually update some measurements at the break of year.
 * the data format requires calculating offset of the "day of year" column, `dayPosition` - this is calculated from averagingPeriod, which is defined in the code - please see the question in the code.


MichalCz:
This is now tested - a small correction of splitting of columns was required.

adapters/arpalazio.js
```diff
@@ -100,89 +96,85 @@ const getParameters = function (averagingPeriod) {
...
-        const fullDate = {
-          utc: date.toDate(),
-          local: date.format()
+export const getStream = function (cityName, url, averagingPeriod, source, orgUrl) {
+  const { metadata } = source;
+  const match = url.match(/[\w]{2}_([\w.]{2,})_([\d]{4})(?:_gg)?.txt/);
+  const parameter = match[1].toLowerCase().replace('.', '');
+  const year = match[2];
+  const unit = getUnit(parameter);
+  const dayPosition = averagingPeriod.value === 1 ? 0 : 1;
```
MichalCz:
```suggestion
    const match = url.match(/[\w]{2}_([\w.]{2,})_([\d]{4})(_gg)?.txt/);
    const parameter = match[1].toLowerCase().replace('.', '');	  const parameter = match[1].toLowerCase().replace('.', '');
    const year = match[2];	  const year = match[2];
    const unit = getUnit(parameter);	  const unit = getUnit(parameter);
    const dayPosition = match[3] === '_gg' ? 0 : 1;
```

MichalCz: @jflasher The above suggestion is for the whole block from L101 to L105, not the single one.

jflasher: @MichalCz if I'm reading this correctly, you're changing how `dayPosition` is calculated? I think is probably fine if it looks good in the resultant data? All of this scraping is a bit hacky anyways.

jflasher:
I think this is good to merge unless is misunderstood the question above.

MichalCz:
This is good to merge. I'll send my suggestion as a separate pull request.

PR #526, opened 2018-10-13
Adapter: Airnow-ke streamed

MichalCz:
This is the streamed version of the Kenyan branch of Airnow. The adapter can be used to fetch all airnow data in one go, but this has not been fully tested.

I'll confirm tomorrow after initial tests - this has just been rebased onto current develop.

### DIFF
--- a/adapters/airnow-ftp.js
+++ b/adapters/airnow-ftp.js
@@ -6,7 +6,7 @@ import { omit, merge, find, filter } from 'lodash';
 import { default as moment } from 'moment-timezone';
 import { convertUnits } from '../lib/utils';
 
-export const name = 'airnow';
+export const name = 'airnow-ftp';
 
 exports.fetchData = function (source, cb) {
   // A workaround to getting rate limited for 6 logins in 1 hr for AirNow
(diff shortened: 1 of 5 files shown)


MichalCz:
@jflasher this version is checked, but it simply grabs all the data ignoring the country. There's a lot of measurements there and there are measurements for Kenya, but also for Canada or the US.

Quite frankly - why not run this adapter as is? They don't use separate files per country - so it's essentially filtering the same list of measurements in each import.

But if that's not so good - I will add cache on the data request, so we won't be fetching the data a couple times.

Let me know what you'd prefer.

jflasher:
I think the intended behavior would be that this adapter is only run once (likely from the us.json source file). We'd remove the ke.json source file and there is no ca.json. So even though there is data from CA, KE and US, the adapter would only get run once. Also, this would be replacing the current `airnow` adapter. Does that make sense?

MichalCz:
Yes, but if I may suggest, I'd look for the source in something like `intl.json`. The same could be used for EEA for example, which could optimize the adapter itself.

But from technical point of view there's no difference whatsoever...

MichalCz:
@jflasher - so what do we do with this one? intl.json or us.json? I'd make the change still today.

jflasher:
intl.json sounds good! I think it’s a bit more clear than having it hidden in a country one.

MichalCz:
@jflasher, just to clear out - the above commit also has the change from ke -> intl.json and changes the active AirNow adapter to the Kenyan HTTP version - so that's good to merge too.

PR #529, opened 2018-10-14
Fix fetches table count reporting on all duplicates.

### LINKED ISSUE #528: New measurement count showing up in fetches table even though new data is not actually inserted
Looking at the fetches table, the total count and count of new items inserted within the `results` object is being incremented even if there is no new data inserted.

You can test by running something like `node index.js -s StateAir_Hanoi` twice in a row. The first time, you should expect to see 24, the second time you would expect 0 (or maybe 1 if you hit the time just right). Instead, 24 is showing up for the `count` and in `results` object in fetches table.

@MichalCz could you look into this? It's not super urgent since I don't believe any new data is being inserted incorrectly, just the counts are off. Thanks!

MichalCz:
Fixes #528

### DIFF
--- a/lib/measurement.js
+++ b/lib/measurement.js
@@ -71,7 +71,7 @@ function createFetchObject (input, source, failures, dryRun) {
       return fetchEnded ? failures : null;
     },
     get count () {
-      return fetchEnded && (counts.inserted || counts.total);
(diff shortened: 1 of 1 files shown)


jflasher:
Checked locally and looks good, thanks!

PR #534, opened 2018-10-19
Fix sources table format by flattening the sources output.

### LINKED ISSUE #533: Form of /sources has changed
Not too many details yet, but I believe something done with the recent fetch updates changed the output form of the `/sources` endpoint. Looks like we may be saving an array of arrays rather than a flat array. 

Fixing the website to handle this right now but we should investigate and fix here if possible.

MichalCz:
Fixes #533 

@jflasher I have limitted possibility of confirming the fix, but indeed the sources are being saved in a "doubled" array. This flattens the array to a single one.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -185,7 +185,12 @@ export async function saveFetches ({timeStarted, timeEnded, itemsInserted, resul
  * @param {Sources} sources
  */
 export async function saveSources (sources) {
-  const inserts = Object.values(sources)
+  const inserts = Object
+    .values(sources)
+    .reduce(
+      (acc, sources) => acc.concat(sources),
+      []
+    )
     .map(data => ({data: JSON.stringify(data)}));
 
   const pg = await getDB();


jflasher:
Thanks @MichalCz , I think this is likely the problem. Would you be able to filter out the dummy source altogether (as it's currently showing up in results) and filter each source down to the standard elements like in below (so we avoid having things like the metadata being saved to the db and showing up in the Arpalazio source when returned via the API)

```
url: "http://discomap.eea.europa.eu/map/fme/latest/",
adapter: "eea-direct",
name: "EEA Andorra",
location: "Andorra Location",
city: "Andorra City",
country: "AD",
description: "Andoran data from UTD service",
sourceURL: "http://www.eea.europa.eu/themes/air/air-quality",
contacts: [
"info@openaq.org"
],
active: true
```

MichalCz:
Yes. I'll send a pull request tomorrow.

PR #550, opened 2018-11-08
Adapter/gios poland

MichalCz:
Hi,

Here's a proposed adapter for handling the data form polish environmental agency, "Główny Inspektorat Ochrony Środowiska", based on this [GIOŚ API Guide](http://powietrze.gios.gov.pl/pjp/content/api).

To test the source run:

`node . -dvs "GIOS"`

All the variables are stored in the source json.

Data is gathered for average per every hour in ug/m3 for all sensor data. The adapter fetches only the data and ignores the AQI endpoint.

The data may change over time (not all data is checked immediately), which is not currently covered by the fetch process - this we could discuss in a separate issue as I think we could do something to periodically revisit older entries and import again only if some checksum differs from the last one.

### DIFF
--- a/sources/pl.json
+++ b/sources/pl.json
@@ -12,5 +12,22 @@
             "info@openaq.org"
         ],
         "active": false
+    },
+    {
+        "url": "http://api.gios.gov.pl/pjp-api/rest",
+        "adapter": "gios-poland",
+        "name": "GIOS",
+        "city": "unused",
+        "country": "PL",
+        "description": "",
+        "resolution": "1 hr",
+        "sourceType": "government",
+        "averagingPeriod": {"unit": "hours", "value": 1},
+        "mobile": false,
+        "attribution": [{"name": "Portal Jakości Powietrza GIOŚ", "url": "http://powietrze.gios.gov.pl/pjp/home"}],
+        "sourceURL": "http://powietrze.gios.gov.pl/pjp/content/api",
+        "contacts": ["info@openaq.org"],
+        "hoursToFetch": 4,
+        "active": false
     }
 ]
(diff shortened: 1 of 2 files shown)


MichalCz:
Hmm... @jflasher I'm not sure why the build failed - it seems to work well for me. Any ideas?

jflasher:
@MichalCz weird. I reran it from the Travis side and it failed on the same test due to a timeout

'should handle measurement errors'

Doesn't look like anything changed there, so not sure why this would be happening.

sources/pl.json
```diff
@@ -12,5 +12,22 @@
...
+        "description": "",
+        "resolution": "1 hr",
+        "sourceType": "government",
+        "averagingPeriod": {"unit": "hours", "value": 1},
+        "mobile": false,
+        "attribution": [{"name": "Portal Jakości Powietrza GIOŚ", "url": "http://powietrze.gios.gov.pl/pjp/home"}],
+        "sourceURL": "http://powietrze.gios.gov.pl/pjp/content/api",
+        "contacts": ["info@openaq.org"],
+        "hoursToFetch": 4,
+        "active": false
```
jflasher: Don't forget to switch to `true`!

MichalCz:
I'll fix this today. :)

sources/pl.json
MichalCz: Done. :)

MichalCz:
Funny thing that I wasn't expecting:

```javascript
    // was:
 // return item.tzDate.add(hoursToFetch, 'hours').isBefore();
    return item.tzDate.isBefore(moment().subtract(hoursToFetch, 'hours'));
```

So in the first version the `item.tzDate` was mutated by `.add` - which means the 4 hours of lookback was added. I expected it to return a new object (like on `String`).

Anyway - now fixed and activated.

MichalCz:
Got it... comitting...

jflasher:
Merging, thanks!

MichalCz:
Yeeeee! I see the data on the website. :) It ain't bad today in Warsaw, I think I'll walk the dog. :)

PR #551, opened 2018-11-11
Hotfix/update scramjet

MichalCz:
Hi,

This is to fix the issue that occurred in the pull request for Polish GIOS adapter. This issue would occur in Scramjet v4.18.16 that would be automatically installed on the next build although should not result in any errors except for the `dummy` test adapter.

In any case this update will fix the issue before it should occur.


jflasher:
Hmm, could you pull against develop? There may be deploy commits that differ between develop and master, but no actual code.

MichalCz:
Done

Issue #547, opened 2018-11-04
Add better filtering for large data sources


jflasher:
If we can add better filtering for ChinaAQIData and Arpalazio, I think it'd go a ways towards avoiding timeouts. My guess is that some of the timeouts (when fetch doesn't finish in time) are happening because we're trying to save a lot of measurements to the database (even if they're duplicates).

MichalCz:
Joe, just thinking out loud, but on arpalazio I think it may be a small change in line 131 - from `filter` to `until`... I'll check that out. Ideally not the whole file would get loaded...

MichalCz:
Well... that doesn't work well with `request` - sadly it crashes because of ending the piped stream before the data is consumed...

I must think of that scenario on newer versions. So this is not really an option. :(

majesticio:
I'm closing this issue because these adapters aren't currently being used and the fetch process has been updated since this issue was created.

PR #573, opened 2018-12-10
Implement error handling on stream pipelines.

MichalCz:
Hi,

As mentioned in #561 here's a solution for crashes occurring in scenarios where some non-scramjet streams are being piped before piping the result into a scramjet stream. The specific case is:

    request(url) -> JSONStream(selector) -> new DataStream.

The error handling between request and JSONStream was not forwarded by default. The solution was to use a new `pipeline` method in scramjet which internally creates pipes between transform streams and forward any errors to the DataStream where it can be handled by `catch`.

The change in `caaqm` adapter may be a possible solution for #547 where similar code with `abort` could be used (which needs to be done, otherwise the unpiped response keeps the process from exiting).

@jflasher please test this on staging - to crash the adapter you can point the url in source to soemthing non-json which caused the crash in previous version.

### DIFF
--- a/adapters/caaqm.js
+++ b/adapters/caaqm.js
@@ -34,19 +34,18 @@ export async function fetchStream (source) {
     body: Buffer.from('{"region":"landing_dashboard"}').toString('base64')
   });
 
+  const requestObject = request(options);
+
   return DataStream
-    .from(
-      () => {
-        const response = request(options);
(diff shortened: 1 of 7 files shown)


jflasher:
Sorry, traveling at the moment but hope to look soon!

PR #575, opened 2018-12-13
Kosovo adapter

tverilytt:
Initial version of Kosovo adapter.


jflasher:
@tverilytt Thanks for getting this started! There are a few things that'll need to be modified to fit our general structure for adapters. A few examples

- We use `node-request` and `cheerio` for requests and parsing respectively. I'd rather keep using those libraries only instead of adding more dependencies unless there is a good reason to do so.

- For debug logging, https://github.com/openaq/openaq-fetch/blob/develop/adapters/eea-direct.js#L19 has an example.

If you want to try and make those changes, please do so. If you want me or someone else to do it, that works too, the parsing logic you've added will be helpful! 

Just let me know whichever works better for you.

tverilytt:
@jflasher Thanks for that. I am happy to look into this. I have used / are using request, but tend to prefer )node)-fetch, since it works similar on both client- and server-side. cheerio I have not used, but if it can parse HTML into JSON, as parse5, it should not be too hard to do the switch I think.

I will have a go and report back :-)
-jo

jflasher:
Thanks @tverilytt, I'll try and look at this soon!

package.json
```diff
@@ -43,7 +43,8 @@
     "tz-lookup": "^6.1.8",
     "winston": "^2.4.4",
     "winston-papertrail": "^1.0.2",
-    "yargs": "^12.0.1"
+    "yargs": "^12.0.1",
+    "node-fetch": "~2.2.x"
```
jflasher: Please drop this requirement as I think it's no longer needed. Also please remove your local changes to package-lock since I think they're also is not needed.

sources/xk.json
```diff
@@ -11,5 +11,18 @@
...
+    {
+        "url": "http://kosovo-airquality.com/secure/ValueTable.html",
+        "adapter": "kosovo",
+        "name": "Kosovo",
+        "city": "",
+        "country": "XK",
+        "description": "Kosovo data from kosovo-airquality.com",
+        "sourceURL": "http://kosovo-airquality.com/secure/index2.html",
+        "contacts": [
+            "info@openaq.org"
```
jflasher: This is a mostly unused field at the moment, but in the future, we may thank contributors or send notifications if they want to know when a source isn't working. With that in mind, feel free to add your email here if you want. Also no worries if not!

adapters/kosovo.js
```diff
@@ -0,0 +1,209 @@
...
+const request = require('request-promise-native');
+const cheerio = require('cheerio');
+
+const log = require('../lib/logger');
+
+const openaq_parameters = ['pm25', 'pm10', 'no2', 'so2', 'o3', 'co', 'bc'];
+
+const averagingPeriod = { unit: 'hours', value: 1 };
+const attribution = [{ name: 'Kosovo AQ', url: 'http://kosovo-airquality.com/secure/index2.html' }];
+const staticData = {
```
jflasher: From view-source:http://kosovo-airquality.com/secure/Kosovo.html, I believe you can get the exact locations if you want to update these based on that.

adapters/kosovo.js
```diff
@@ -0,0 +1,209 @@
+/**
+ * This code is responsible for implementing all methods related to fetching
+ * and returning data for a Kosovo sources (HTML page)
+ */
+'use strict';
+
+const moment = require('moment-timezone');
+const request = require('request-promise-native');
```
jflasher: Please follow a similar pattern to https://github.com/openaq/openaq-fetch/blob/develop/adapters/airnow-http.js#L11 and include the `REQUEST_TIMEOUT` so that the adapter doesn't hang the entire process. Looks like we lost this for the adapters using `request-promise-native`, I'll fix these separately.

adapters/kosovo.js
```diff
@@ -0,0 +1,209 @@
...
+const cheerio = require('cheerio');
+
+const log = require('../lib/logger');
+
+const openaq_parameters = ['pm25', 'pm10', 'no2', 'so2', 'o3', 'co', 'bc'];
```
jflasher: Please import these similar to https://github.com/openaq/openaq-fetch/blob/develop/adapters/airnow-http.js#L5

adapters/kosovo.js
```diff
@@ -0,0 +1,209 @@
...
+  return new Promise(function(resolve, reject) {
+  return request(url)
+    .then(html => resolve(html))
+    .catch(error => reject(error));
+  });
+}
+
+function getParameterAndUnit(header) {
+  const parunit = header.split('[');
+// header.substring(0, header.indexOf('['))
```
jflasher: Please just remove any commented out code if it's no longer needed.

adapters/kosovo.js
```diff
@@ -0,0 +1,209 @@
...
+      log.debug('--------------- raw JSON table -------------');
+      log.debug(rawTable);
+      const table = getTable(rawTable).filter(measurement => openaq_parameters.includes(measurement.parameter));
+      resolve({ name : module.exports.name, measurements : table });
+    })
+    .catch(error => reject(error));
+  });
+}
+
+module.exports.name = 'kosovo';
```
jflasher: Can you please move this up top to right below the imports to keep it consistent with other adapters?

adapters/kosovo.js
```diff
@@ -0,0 +1,209 @@
+/**
```
jflasher: Ok, so something really weird. The map and table data do not match! I verified we're getting the proper data for the table and I think we should go with that, but can you just add a comment in the header here that as of this date the two pages do not agree? I would like to be as transparent as possible about that.

adapters/kosovo.js
tverilytt: Cool, ok to hard-code the locations from that page? I assume they are fairly static...;-)

jflasher: Yep!

package.json
tverilytt: Hm, think this was already removed in last push. Double-checked now, should be gone. Will try push new version this weekend.

adapters/kosovo.js
tverilytt: Hm, just checked one station now (Drenas(, and seems the PM10 and PM2.5 are unequal, whilst NO2, SO2 and CO match. NO does not match, but that parameter is not included by Open AQ it seems. I make a note :-)

tverilytt:
@jflasher  Pushed new version, hopefully addressing your feedback.

The "npm run lint" gives a load of errors on spaces and indentation, do you use some auto-formatter for fixing this?

I want to do some more updates, but mainly regarding cheerio, trying to use it a bit more intelligently, (like I started doing in my version of the Thailand adapter...), but the resulting measurements should stay the same.

Cheers
-jo

tverilytt:
@jflasher Cool, thanks!

I found the [js-beautify](https://github.com/beautify-web/js-beautify) package which with this command resolved most complaints:
js-beautify --indent-size 2 --space-after-named-function --space-after-anon-function --replace kosovo.js 

Removed package-lock.json, but still shows branch conflict, maybe it just takes some time ?

Cheers
-jo

jflasher:
Sorry, this got lost as I was on the road for a bit. Merging in now!

tverilytt:
> Sorry, this got lost as I was on the road for a bit. Merging in now!

Cool, thanks, I can see the data through the API :-)

PR #582, opened 2018-12-19
Fix pipelining errors

MichalCz:
This is a retry of #573:

As mentioned in #561 here's a solution for crashes occurring in scenarios where some non-scramjet streams are being piped before piping the result into a scramjet stream. The specific case is:

    request(url) -> JSONStream(selector) -> new DataStream.

The error handling between request and JSONStream was not forwarded by default. The solution was to use a new pipeline method in scramjet which internally creates pipes between transform streams and forward any errors to the DataStream where it can be handled by catch.

The change in caaqm adapter may be a possible solution for #547 where similar code with abort could be used (which needs to be done, otherwise the unpiped response keeps the process from exiting).

As to the addition to the previous changes, a small change in returning an array from `DataStream..use` in [db.js:164](https://github.com/openaq/openaq-fetch/commit/07335d2d19a95d9d9869b8f4bca45a3f1482f78c#diff-be228df81b99a384aa50ad50bde7fca0R164) - this is due to the fact that prior to scramjet-core 4.17 the outcome of `use` was not checked as it should and the output could be anything, after a fix it's converted to a stream. The fix here is returning an empty array which results in an empty, but ended stream which in turn is converted to a promise.

### DIFF
--- a/adapters/caaqm.js
+++ b/adapters/caaqm.js
@@ -34,11 +34,18 @@ export async function fetchStream (source) {
     body: Buffer.from('{"region":"landing_dashboard"}').toString('base64')
   });
 
+  const requestObject = request(options);
+
   return DataStream
-    .from(
-      request(options)
-        .pipe(JSONStream.parse('map.station_list.*'))
+    .pipeline(
+      requestObject,
+      JSONStream.parse('map.station_list.*')
     )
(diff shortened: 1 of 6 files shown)


jflasher:
Superseded by #583

Issue #557, opened 2018-11-21
caaqm adaptor is not working since 19 November 2018


nishadhka:
![No updates](https://user-images.githubusercontent.com/3928997/48823335-74749c80-ed86-11e8-82e9-bab6f9c99eb2.png)

There is a change in urls and data stream output for this adaptor
**urls**
1. the current source url (in source/in.json) is invalid ```https://app.cpcbccr.com/caaqms/caaqms_landing_map_all```, the url ```https://app.cpcbccr.com/caaqms/caaqms_landing_map``` to be alternative, showing 405.
1. Adaptor url for specific station is changed from ```https://app.cpcbccr.com/caaqms/caaqms_view_data``` into ```https://app.cpcbccr.com/caaqms/caaqms_view_data_v2```

**data stream output**

1. The source url console output json is no longed gives parameter list, thus gives error in adaptors caaqm.js in lines 
```
        // Check if it has parameters we want
        for (var i = 0; i < site['parameter_status'].length; i++) {
          const p = site['parameter_status'][i];
          if (['PM2.5', 'PM10', 'NO2', 'CO', 'SO2', 'NO2', 'BC', 'Ozone'].includes(p['parameter_name'])) {
            return siteStream.whenWrote({
              station_id: site['station_id'],
              station_name: site['station_name'],
              coords: {latitude: Number(site['latitude']), longitude: Number(site['longitude'])}
            });
          }
        }
      },
```

nishadhka:
this patch is collecting data https://github.com/nishadhka/openaq-fetch/tree/caaqm, however only NCR region 51 station rather than 136 stations

jflasher:
Just a note that the page was loading, but now isn't and it exposed an error in the code. So we need to fix that up before we can reactivate the source. But it should be live again in the near future.

PR #586, opened 2018-12-27
CAAQM- all stations fetch url

nishadhka:
For all stations fetch, the url for caaqm has to be https://app.cpcbccr.com/caaqms/caaqms_landing_map_all instead of https://app.cpcbccr.com/caaqms/caaqms_landing_map


jflasher:
@nishadhka Sorry, missed this until now. Do you know what the difference between these two urls is? We changed because of https://github.com/openaq/openaq-fetch/issues/557#issue-382978017 but maybe we are using the wrong url?

jflasher:
Thanks! Merging this now.

PR #617, opened 2019-06-20
Reenabling China source

### LINKED ISSUE #614: Reenable China adapter
We were seeing a lot of load on the database due to large amount of inserts and the database ran out of space. We upped the space and have disabled the Chinese adapter temporarily to see if that was the issue. Once we confirm fetch is working again properly, we should reenable Chinese adapter and see what happens.

jflasher:
closes #614

### DIFF
--- a/sources/cn.json
+++ b/sources/cn.json
@@ -81,7 +81,7 @@
         "contacts": [
             "info@openaq.org"
         ],
-        "active": false
+        "active": true
     },
     {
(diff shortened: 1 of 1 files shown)


pjrobertson:
It appears the China adapter is disabled and MEP data cannot be sources. Is there a reason for this?

A reliable repository of MEP data is here: https://github.com/HeQinWill/CNEMC/

PR #621, opened 2019-07-15
add metadata migrations

sethvincent:
This adds the metadata migration files introduced in this openaq-api pr: https://github.com/openaq/openaq-api/pull/403

@jflasher Let me know if this PR needs anything else.


jflasher:
@sethvincent sorry for the delay here. If I am reading this right, it looks like there is 1 view creation, 1 table creation and then an edit to both the table and the view. Would it be possible to collapse these down to just 2 migration files, 1 for the table create and 1 for the view create? Basically, if nothing has hit the staging db yet (which I don't think it has), I'd prefer to have fewer migrations. The only thing this will screw up is your own local db, you'll probably need to drop the database and remake it or remove the table/view and entries from the migrations table so you can retest applying the new migrations. Let me know if that approach doesn't make sense.

sethvincent:
@jflasher that makes sense. Should we reduce down to 1 migration file in the api repo as well?

sethvincent:
@jflasher updated to reduce the migrations down to one file.

jflasher:
Merging and then I'll run a fetch which should create the tables. You should then be good to go to merge openaq-api once that's done.

jflasher:
@sethvincent fetch task has run on staging and confirmed the new table is present (though empty).

PR #622, opened 2019-07-23
fix location metadata migration

sethvincent:
This may close https://github.com/openaq/openaq-api/issues/406


jflasher:
Merging and will drop the existing table and migration record so this update migration will run when triggered via fetch.

PR #637, opened 2019-12-08
update au/qld and au/nsw attribution

andrewharvey:
The requested attribution for both of these sources has changed, per their websites.


andrewharvey:
They change their names every election :wink: . I've also updated the URL based on the redirect.

PR #638, opened 2019-12-08
au/qld filter null island locations

andrewharvey:
the au/qld feed contains two locations with null island coordinates. I've reported this upstream but in the meantime it seems sensible to filter these locations out.

### DIFF
--- a/adapters/queensland.js
+++ b/adapters/queensland.js
@@ -66,6 +66,12 @@ const getParameterUnit = function (parameter) {
   }
 };
 
+// hardcoded coordinates for locations with no coordinates from source
+var locationCoordinates = {
+  'Miles Airport':  { latitude: -26.8088, longitude: 150.1799 },
+  'Hopeland': { latitude: -26.8637, longitude: 150.5028 }
+};
+
 var formatData = function (data, source) {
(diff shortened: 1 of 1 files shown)


sruti:
Is there any way to get the coordinates from somewhere else and hardcode them?

andrewharvey:
> Is there any way to get the coordinates from somewhere else and hardcode them?

- The application at https://apps.des.qld.gov.au/air-quality/ excludes these two locations from the map.
- The station details at https://apps.des.qld.gov.au/air-quality/stations/?station=hop doesn't reveal anything.
- https://www.researchgate.net/figure/Approximate-location-of-Hopeland-air-quality-station-shown-by-blue-circle-Station-is-on_fig8_326224937 indicates an approximate location but I don't know if we can or should copy it from there and if an approximate location is okay?
- https://www.researchgate.net/figure/The-Miles-Airport-measurement-location-green-pin-red-lines-are-major-roads-purple_fig9_326224937 also has an exact location, but again is it okay to copy this location?

If okay, I'll take it from the last two links, just taking two locations from the map is probably fine.

andrewharvey:
> Is there any way to get the coordinates from somewhere else and hardcode them?

@sruti okay I've done that now.

PR #643, opened 2019-12-11
add license as a source property

andrewharvey:
I believe it's important to keep track of what license source data is provided under, so I'm proposing to add a new `license` property to the source object, using the standard SPDX identifier https://spdx.org/licenses/.

### DIFF
--- a/docs/source.md
+++ b/docs/source.md
@@ -33,6 +33,7 @@ Optional properties (used to add information to measurements):
 * `city (String)` - City name
 * `location (String)` - Name of the location
 * `mobile (Boolean)` - Is the source mobile
+* `license (String` - The license for the data specified using a [SPDX identifier](https://spdx.org/licenses/).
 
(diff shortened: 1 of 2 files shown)


sruti:
@RocketD0g may have some thoughts on this

andrewharvey:
The only thing SPDX doesn't capture well is public domain (eg US federal works?), or custom licenses ideally we'd be able to just link to the URL for the custom license, and somehow capture which sources are public domain, while still being able to capture all the licenses like CC0, CC-BY (and all the CC versions and variants).

andrewharvey:
Any more thoughts on if this will be a useful addition? I still think it's useful to help keep track of all contributing source licensees.

RocketD0g:
Hey @andrewharvey - Thanks again for this question! This was an issue that falls in my wheelhouse, and your question ended provoking a lot of thoughtful discussions for us and some tweaks in our policies for how we want to approach these kinds of things. That said, I'm just heading out on Parental leave until April and will need to hold on this question for the time being.

PR #648, opened 2019-12-12
fix bug parsing NSW, Australia feed

### LINKED ISSUE #646: Australia high SO2 (potential bug?)
OpenAQ reports very high SO2 near Sydney (~300ppb) while the governmental website shows very low concentrations. Not sure whether there is something wrong or not. 

Screen Shot 2019-12-11 at 11 20 35 AM
Screen Shot 2019-12-11 at 11 20 17 AM

andrewharvey:
closes #646, refering to the diagram from the issue, rows which didn't have a region column were incorrectly indexed resulting in values and parameters being mixed up, for example "Randwick", "Richmond" were broken but "Cook And Phillip" and "Parramatta North" were working.

cc @ReaRuiRay

### DIFF
--- a/adapters/nsw.js
+++ b/adapters/nsw.js
@@ -10,6 +10,18 @@ const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
 
 export const name = 'nsw';
 
+/*
+ * There are two pages for live air quality data
+ *
+ * Index Values:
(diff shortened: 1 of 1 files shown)


andrewharvey:
@sruti anything else I can do to help get this one in? It's a bit more urgent than the other PRs since we currently have wrong values being reported.

sruti:
Looks good to me! 👍

Issue #552, opened 2018-11-12
New data source for Iceland 


nishadhka:
The data source is https://www.ust.is/einstaklingar/loftgaedi/maelingar/ . There are about 26 stations giving real time data with most of [openaq-data-fromat metadata](https://github.com/openaq/openaq-data-format) details as well.

RocketD0g:
Great, thanks for the heads up on this! I am putting a tag to put this issue on hold.

RocketD0g:
Great - thanks so much for the heads up on this!

RocketD0g:
Hi @Stalfur  - any update on this? We'd love to add Icelandic data.

Stalfur:
Hi @RocketD0g - yes we are now rolling out an API, I'll keep this updated, should be up within a month.

Stalfur:
https://api.ust.is/aq is now live - we start with station info and hourly values for past 24 hours.

Stalfur:
If there is anything that would be better for OpenAQ then we can accommodate that probably

RocketD0g:
Wow, awesome @Stalfur! This is great! **Thanks for sharing this!** It looks like all the metadata we need on our end is in there, but @sruti will be the one to make sure I'm not speaking too quickly.

Two small clarifications: 

(1) I'm assuming since the data are shared hourly that these are hourly-averaged data? (We've had a couple instances where data are shared hourly but the time-average on a given measurement itself is shorter, so just double checking).

(2) I also assume the time is reported in GMT, UTC+0, the timezone for Iceland as I understand it, but also just double-checking.

Lastly, @Stalfur - would you be open for us potentially chatting with you via email about the creation of this API? We would love to amplify this great example of sharing key air quality data in a programmatic and open way! If so, please shoot me an email at christa@openaq.org (or feel free to leave yours here and I will reach out to you, but I realize that your email would then be shared publicly).

Thanks again - this is fantastic!

Stalfur:
Hello @RocketD0g 

(1) Yes these are hourly averaged data. We will add to the clarification, adding units and averaged time.

(2) Yes luckily being at UTC+0 allows us to pay no mind to timezone conversions for now.

Lastly yes I'd be happy to, I'll drop you an e-mail from my work email.

olafveerman:
A first version for the adapter is implemented in the Pull Request above.

Thanks @Stalfur for providing the data through this API, it was a breeze to get it in.

[...]

PR #659, opened 2020-01-03
Add adapter for Icelandic data

olafveerman:
This is a first draft for an adapter to pull in the Icelandic data reported in #552. It's fairly functional, though the following is missing:

* [x] get coordinates from the stations endpoint (https://api.ust.is/aq)
* [x] add the units 
* [x] decide who the contact is
* [x] figure out what we want to do with the city property. The dry-run currently fails because city is an empty string.  
@sruti @jflasher now that the coordinates are mandatory, should we loosen the requirements on the city attribute?

### DIFF
--- a/adapters/ust-ist.js
+++ b/adapters/ust-ist.js
@@ -63,7 +63,7 @@ export async function fetchData (source, cb) {
 
 function parseParams (params) {
   // Array with the valid parameters in the object
-  const validParams = Object.keys(params).filter(p => acceptableParameters.includes(p.toLowerCase()));
+  const validParams = Object.keys(params).filter(p => acceptableParameters.includes(p.toLowerCase().replace('.', '')));
 
   return validParams.map(p => {
     // Assumes that '0' is always latest
@@ -81,7 +81,7 @@ function parseParams (params) {
         utc: date.toDate(), // 2020-01-03T04:00:00.000Z
         local: date.format('YYYY-MM-DDTHH:mm:ssZ') // '2020-01-03T04:00:00+00:00'
       },
-      parameter: p.toLowerCase(),
+      parameter: p.toLowerCase().replace('.', ''),
       value: Number(latestM.value),
       unit: params[p].unit,
       averagingPeriod: resolution
--- a/adapters/ust-ist.js
+++ b/adapters/ust-ist.js
@@ -1,0 +1,81 @@
+'use strict';
+
+import { default as moment } from 'moment-timezone';
+import { acceptableParameters, promiseRequest } from '../lib/utils';
+
+export const name = 'ust-ist';
+
+export async function fetchData (source, cb) {
+  try {
+    const data = JSON.parse(await promiseRequest(source.url));
+
+    // Generate an array of station ID
+    const stations = Object.keys(data);
+
+    const measurements = stations.reduce((acc, stationId) => {
+      const station = data[stationId];
+
+      // Only interested in 1 hour resolution.
+      if (station.resolution !== '1h') return acc;
+
+      const baseMeta = {
+        location: station.name,
+        city: '',
+        averagingPeriod: { value: 1, unit: 'hours' },
+        coordinates: { latitude: 0, longitude: 0 }
+      };
+
+      const latestMeasurements = parseParams(station.parameters);
+
+      return acc.concat(latestMeasurements
+        .map(m => ({ ...baseMeta, ...m }))
+      );
+    }, []);
+    cb(null, {name: 'unused', measurements});
+  } catch (e) {
+    cb(e);
+  }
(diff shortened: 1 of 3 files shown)


Stalfur:
municipality should be equal to city

olafveerman:
Thanks @Stalfur, added the municipality as the city.
Do you have any insight into the units of each parameter? Are they the units listed in this table? https://github.com/openaq/openaq-data-format#preferred-units

RocketD0g:
@Stalfur, thanks again for sharing this API! I'm not sure if [this page](https://loftgaedi.is/?zoomLevel=7&lat=64.894972&lng=-18.675028) shows the same data, but if so, is it correct to say for the units: 

PM10  - ug/m3
PM2.5 - ug/m3
NO2 - ug/m3
SO2 - ug/m3

Stalfur:
Yes they are, I have now added unit to the JSON within each parameter - I hope it doesn't break as now the values get a sequential number automatically in JSON generation.

Stalfur:
I've made another change, I moved resolution within each parameter. @olafveerman @RocketD0g

olafveerman:
The last commit accounts for the latest changes by @Stalfur.

With the unit and averaging period, the dry-run passes without an issue. Marking this as Ready for Review.

adapters/ust-ist.js
```diff
@@ -0,0 +1,90 @@
...
+ *   }
+ * })
+ *
+ * @returns [ { value: 0.154717, date: 2020-01-03 03:00:00. parameter: 'no2' }]
+ *
+ */
+
+function parseParams (params) {
+  // Array with the valid parameters in the object
+  const validParams = Object.keys(params).filter(p => acceptableParameters.includes(p.toLowerCase()));
```
sruti: This is leaving out PM2.5 since the acceptable parameter looks like 'pm25' and this makes it 'pm2.5'

adapters/ust-ist.js
```diff
@@ -0,0 +1,90 @@
...
+    return {
+      date: {
+        utc: date.toDate(), // 2020-01-03T04:00:00.000Z
+        local: date.format('YYYY-MM-DDTHH:mm:ssZ') // '2020-01-03T04:00:00+00:00'
+      },
+      parameter: p.toLowerCase(),
+      value: Number(latestM.value),
+      unit: params[p].unit,
+      averagingPeriod: resolution
+    };
```
sruti: Missing 'attribution' object in the measurement like so: `["name": "Umhverfisstofnun", "url": "https://ust.is" ]` (or however they would like to be attributed)

sources/is.json
```diff
@@ -0,0 +1,14 @@
+[
+    {
+        "url": "https://api.ust.is/aq/a/getLatest",
+        "adapter": "ust-ist",
+        "name": "Umhverfisstofnun",
+        "country": "IS",
+        "description": "Environment Agency of Iceland",
+        "sourceURL": "https://ust.is",
+        "contacts": [
```
sruti: @Stalfur do you mind being the contact or know someone who should be? If not, we can put down `info@openaq.org`

Stalfur: We have a loftgaedi@umhverfisstofnun.is (shared mailbox) that should be used.

adapters/ust-ist.js
Stalfur: This is perfect attribution

Stalfur: I've been wondering if it would be "more correct" to move the values into a values {} within the parameter, thoughts?

sruti: In your API? What would that look like? I'm not quite sure I understand

sources/is.json
olafveerman: Great, was added. Thanks

adapters/ust-ist.js
olafveerman: Fixed in the latest commit.

adapters/ust-ist.js
olafveerman: I was under the impression that the adapter inherits the attribution from the source. That's not the case?

sruti: @olafveerman No, it looks like attribution doesn't [automatically inherit](https://github.com/openaq/openaq-fetch/blob/develop/lib/measurement.js#L139) from the source file like sourceName and sourceType do. I think it's because some adapters (like Airnow) get the attribution dynamically from the source. Or they're hardcoded. But maybe this could be changed so attribution is inherited from the source file if it's not set in the adapter.

sruti: Didn't add automatic inheritance for the attribution because there are cases where the source name and desired attribution don't match up so it's better to be explicit.

PR #697, opened 2020-03-30
Fixed arpae PM2.5 parameter to align with acceptedParameters value

AJExtravaganza:
arpae-parameters.js currently parses a parameter of 'pm2.5' for PM2.5 instead of 'pm25', preventing ingestion of PM2.5 measurements from arpae adapter.

### DIFF
--- a/adapters/arpae.js
+++ b/adapters/arpae.js
@@ -83,7 +83,7 @@ const parameters = {
   '21': { parameter: 'c6h5-ch3', unit: 'ug/m3' },
   '38': { parameter: 'no', unit: 'ug/m3' },
   '82': { parameter: 'o-xylene', unit: 'ug/m3' },
-  '111': { parameter: 'pm2.5', unit: 'ug/m3' }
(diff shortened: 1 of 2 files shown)


sruti:
Thanks for this @AJExtravaganza! Do you know if data from the arpae source feeds into the EEA data? It looks like Italy's data is [available through the EEA](http://discomap.eea.europa.eu/map/fme/AirQualityUTDExport.htm) and wondering if there may be conflicts/duplicates if we enable Italy in the EEA adapter and also fix the arpae adapter?

Also, could you merge into the `develop` branch please? We usually merge changes into `develop` first, check nothing broke, then merge into `master`.

sruti:
Nvm I looked into it and there doesn't seem to be an overlap. And even so, our system should be able to handle it. If you can change it so it merges into develop, it's good to go 👍

AJExtravaganza:
Whoops, sorry @sruti! Have opened #699 targeting correct branch.

PR #707, opened 2020-04-17
Adding remaining EEA-direct sources

### LINKED ISSUE #677: Several countries in EEA we are not pulling
Some **real** low-hanging fruit: Looks like there are several countries we are not pulling from, from EEA [(via direct download option)](http://discomap.eea.europa.eu/map/fme/AirQualityUTDExport.htm):

- Bosnia - BA
- Bulgaria* - BG 
- Cyprus - CY
- Denmark* - DK
- Estonia - EE
- Georgia - GE
- Greece - GR
- Romania - RO
- Sweden - (Which we have previously ingested, but is a broken adapter currently)

*It currently _looks_ like we have Bulgarian data and Denmark data, but this is due to an issue with the AirNow data source mislabelling a two-letter country code.

magsyg:
This push has made these sources pull from EEA
- Bulgaria
- Cyprus
- Georgia
- Greece
- Romania

Closes #677


magsyg:
I checked out the other EEA sources for countries which already have sources. EEA for Iceland, Great Britain, Norway, Netherlands and Slovenia seemed to be complete duplicates of the already existing sources. Bosnia's EEA source did not seem to work, The EEA sources of Italy and Poland seemed to have some of the same values, but it looked like the EEA sources had a lot of more stations than the existing sources. Sweden's source did not load anything, could be a problem with the adapter, but the EEA source does seem to provde a lot more stations than the existing swedish source, which only has airdata from Stockholm, not the rest of sweden

PR #716, opened 2020-04-22
Added EEA Sources for Poland, Sweden and Italy

### LINKED ISSUE #677: Several countries in EEA we are not pulling

magsyg:
Added EEA Sources which had an abundance of extra stations compared to the country's existing sources, these are:
- Poland
- Italy
- Sweden

closes #677

### DIFF
--- a/sources/it.json
+++ b/sources/it.json
@@ -80,7 +80,7 @@
     "contacts": [
       "info@openaq.org"
     ],
-    "active": true
+    "active": false
   },
   {
     "url": "http://discomap.eea.europa.eu/map/fme/latest/",
--- a/sources/it.json
+++ b/sources/it.json
@@ -81,5 +81,18 @@
       "info@openaq.org"
     ],
     "active": true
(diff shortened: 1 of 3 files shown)


espenairmine:
Need to wait with this PR till the following PR's have been approved, as we need to update filtering for EEA for Italy:
#366 
#720 
#721

PR #728, opened 2020-05-07
Created an adapter for Singapore

### LINKED ISSUE #12: Singapore - Sources (not useable yet)
Listed a priority given recent AQ issues in area but NOTE: There is an issue with these data as they are currently presented: They are not listed as individual stations, but rather as aggregated stations that represent different regions of the city. WRI/GFW is contacting NEA on our behalf to see if the station-level data are accessible. We will not proceed with these data until we received the station-level information.

After discussions with WRI/GFW, they are particularly interested in AQ data from countries affected by the fires in Indonesia.

AQ data for Singapore:
http://www.nea.gov.sg/anti-pollution-radiation-protection/air-pollution-control/psi/pollutant-concentrations/type/PM25-1Hr#pollutant

If anyone knows differently, please comment.

magsyg:
The data is gathered from https://www.haze.gov.sg/,
Has 5 stations, and parameters for pm25, pm10, so2, co, o3 and no2.
All parameters except pm25 has to be converted from an PSI, and the convertion formula found here: 
https://www.haze.gov.sg/docs/default-source/faq/computation-of-the-pollutant-standards-index-(psi).pdf

Updates hourly

Should close #12

### DIFF
--- a/adapters/singapore.js
+++ b/adapters/singapore.js
@@ -1,0 +1,192 @@
+/**
+ * This code is responsible for implementing all methods related to fetching
+ * and returning data from Singapore
+ * 
+ * data.
+ */
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import _ from 'lodash';
+import { default as moment } from 'moment-timezone';
+import { default as rp } from 'request-promise-native';
+
+const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+
+exports.name = 'singapore';
+
+/**
+ * Fetches the data for a given source and returns an appropriate object
+ * @param {object} source A valid source object
+ * @param {function} cb A callback of the form cb(err, data)
+ */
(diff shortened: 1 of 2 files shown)


RocketD0g:
Thanks so much for this, @magsyg (and actually all of the work you've been doing on here!!).

Unfortunately, we've been in contact with Singapore's NEA and [these data are actually not station-level,](https://github.com/openaq/openaq-fetch/issues/12#issuecomment-316458715) but the average of several stations in each region (North, East, West, South, Central). Since it's not station level, we can't ingest it into the system.

As a secondary issue for the non-PM2.5 data, we also, to date, have been very very reticent to take in countries' AQI/PSI values and back-calculate out the physical data. AQI/PSIs tend to change over time with a given country's policies, but if we hard code the PSI formula in (which is the only way to really do it), we don't have a reliable way to know when this inevitably changes and we're reporting out incorrect values that we've generated from their source data.

sruti:
Closing this since we can't ingest the data at this time

PR #734, opened 2020-05-19
Created an adapter to fetch PM values from Denmark

magsyg:
This is an adapter to fetch the 4 pm stations in denmark, which updates hourly from this site:
https://envs2.au.dk/Luftdata/Presentation/Graph/PM.

The current source EEA, does not provide us any PM values for Denmark, so this should add the data we lack for Denmark. 

Only problem with this source is that the timestamp is currently a bit shaky, because it is not given in the sourcecode, or atleast nowhere I could find, but the site displays hourly data, so I had to do a workaround where the timestamp is based around nearest hour, which does seem to correspond correctly to what is displayed on the site.

### DIFF
--- a/adapters/denmark-pm.js
+++ b/adapters/denmark-pm.js
@@ -1,0 +1,101 @@
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import _ from 'lodash';
+import { default as moment } from 'moment-timezone';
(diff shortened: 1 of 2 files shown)


sruti:
Because this is not a governmental source, we reached out to them to 1) understand what sensors they're using and 2) if they would be comfortable with sharing data on the platform (in the past, the research-grade data we've added has been at the request of the researchers). So putting this on hold until we hear back

PR #746, opened 2020-06-03
Fixed the turkish adapter

### LINKED ISSUE #19: Turkey - Sources
Map of stations with coordinates and current readings and stations' current data are here (though not on unique urls):

http://www.havaizleme.gov.tr/Default.ltr.aspx

'TÜM İSTASYONLAR' = all stations

Clicking on the stations reveals site coordinates (click 'station description) and pollutant types measured.

(Sidenote: They use an AQI system with breakpoints same as the US EPA)

magsyg:
Fixed the turkish adapter, this should pull ca 350 sources hourly

Should close #19

### DIFF
--- a/adapters/turkey.js
+++ b/adapters/turkey.js
@@ -1,6 +1,6 @@
 /**
  * This code is responsible for implementing all methods related to fetching
- * and returning data for the Taiwanese data sources.
+ * and returning data for the Turkish data sources.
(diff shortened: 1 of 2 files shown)


sruti (changes requested):
Some measurements are missing. But otherwise works perfectly!

PR #747, opened 2020-06-04
Adapter for Air4Thai

### LINKED ISSUE #664: New Thai data source
A colleague shared this data source - including PM2.5 for Thailand): http://air4thai.pcd.go.th/webV2/history/

(A reminder that a previous colleague in #542 also shared this one: http://aqmthai.com/index.php?lang=en )

magsyg:
Created an adapter based on the issue from #664, it looks like it has a lot of more stations and datatypes compared to the old Thailand source. It looks like this source overlaps with the old Thailand source, but this one has a lot of more data, so the old one should be replaced with this one.

Should close #664


adapters/air4thai.js
```diff
@@ -0,0 +1,101 @@
...
+      date: {
+        utc: dateMoment.toDate(),
+        local: dateMoment.format()
+      },
+      coordinates: {
+        latitude: Number(item.lat),
+        longitude: Number(item.long)
+      },
+      attribution: [{name: 'Air4Thai', url: 'http://air4thai.pcd.go.th/webV2/'}],
+      averagingPeriod: {unit: 'hours', value: 1}
```
sruti, line 74: It actually looks like the averaging periods are different for different pollutants, according to the labels on this page: http://air4thai.pcd.go.th/webV2/region.php?region=0 and this info page in Thai (though you can infer from the numbers what's meant): http://air4thai.pcd.go.th/webV2/aqi_info.php

PM2.5 and PM10 are 24hr averages, O3 and CO are 8 hr averages, NO2 and SO2 are 1 hr averages

sources/th.json
```diff
@@ -12,5 +12,19 @@
...
         ],
         "active": true
+    },
+    {
+        "url": "http://air4thai.pcd.go.th/forappV2/getAQI_JSON.php",
+        "adapter": "air4thai",
+        "name": "Air4Thai",
+        "city": "",
+        "country": "TH",
+        "description": "",
```
sruti, line 22: I would add the agency name in description so it's there "Air Quality and Noise Management Bureau, Pollution Control Board"

sruti (changes requested):
Some small changes, otherwise looks good to go @magsyg! Thanks so much!

PR #753, opened 2020-06-22
Fetcher/montenegro

### LINKED ISSUE #742: New Data: Montenegro
Found this source for Montenegro, which does not seem to be in our database at the moment:
http://www.epa.org.me/vazduh/
Looks to be data based on stations, and is based on raw measurements.

magsyg:
Created a fetcher for Montenegro, everything seems to be in order with this source, stationary, raw and hourly, and it is not in EEA

Should close #742


russbiggs:
Thanks for this @magsyg

PR #754, opened 2020-06-22
Created a fetcher for Trinidad and Tobago

### LINKED ISSUE #744: New Data: Trinidad and Tobago
Found this source for Trinidad and Tobago, does look like it is stationbased and raw:
http://ei.weblakes.com/RTTPublic/DshBrdStation

magsyg:
Created a fetcher for Trinidad and Tobago. The Source is kind of small, only has around 3 stations, but it seems that only 2 of them are reporting at the moment. Data is raw and hourly

Should close #744


russbiggs:
@majesticio this seems to still be a working data source, worth investigating

majesticio:
resolved by #949

PR #755, opened 2020-06-23
Created a new fetcher for Mexico

### LINKED ISSUE #672: Mexican AQ data
Hourly data can be found here: https://sinaica.inecc.gob.mx/ by clicking on 'Modulos' then 'Datos crudos' and then choosing the station. Requires scraping but looks straightforward.
 
To verify/investigate:
- [ ] hourly-averaged
- [ ] station coordinates
- [ ] easier, programmatic way to access the data
- [ ] doesn't duplicate existing Mexican data from AirNow

magsyg:
Based on the issue #672, data is raw, stationary and hourly.

The fetcher can have a small problem with pulling at times, because of "Error: ESOCKETTIMEDOUT", which may come from there are to many requests running at the same time, around 160.
I could not find any way to avoid this error, a little help would be very welcome.

Does seem to overlap a bit with the data from Stateair, but this source is much more stable than stateair, and has stations that currently are inactive on stateair, so the mexican data from stateair should be disabled

Should close #672


adapters/mexico.js
```diff
@@ -0,0 +1,195 @@
...
+          .map(script => $(script).html())
+          .filter(script => script.search('conts = {') !== -1)[0];
+
+        // Metadata from the documentscript, turns it into json to get latitude, longitude and location
+        var meta = values.substring(values.indexOf('cump = {'));
+        meta = meta.substring(String('cump = {').length - 1, meta.indexOf('};') + 1);
+        meta = JSON.parse(meta);
+        template['coordinates'] = {
+          latitude: Number(meta.lat),
+          longitude: Number(meta.lat)
```
sruti, line 157: This should be `meta.lon`

adapters/mexico.js
```diff
@@ -0,0 +1,195 @@
...
+        // formats the data and values from the documentscript into readable data
+        var data = values.substring(values.indexOf('conts = {'));
+        data = data.substring(String('conts = {').length - 1, data.indexOf('};') + 1);
+        data = JSON.parse(data);
+        // Loops through all parameters of the site
+        Object.values(data).forEach(param => {
+          // Loops through all the measurements for each parameter
+          param.forEach(d => {
+            if (d != null) {
+              const dateMoment = moment.tz(d.fecha + ' ' + d.hora, 'YYYY-MM-DD H', 'America/Mexico_City');
```
sruti, line 170: It looks like some cities are in a different timezone to Mexico City, for example [Chihuahua ](https://sinaica.inecc.gob.mx/estacion.php?estId=428) is in UCT-8 (or UTC-7 in the summer) while most other cities are UTC-6. Hopefully it should only be a few cities where this is the case but it does list the timezone after `Zona horaria:`

adapters/mexico.js
```diff
@@ -0,0 +1,195 @@
...
+      if (found) break;
+    }
+  };
+  var measurements = [];
+  // Loops through each oage
+  pages.forEach(page => {
+    var $ = cheerio.load(page);
+    // base template
+    var template = {
+      attribution: [{name: 'Sinaica.inecc', url: 'https://sinaica.inecc.gob.mx/index.php'}],
```
sruti, line 132: I would simplify the name to SINAICA

sruti (changes requested):
This works great! It throws the error `unable to load data urls` about half the time, but I think it's okay since we run fetch every 10 minutes and when it does get data, it gets data for the past ~16 hours. A few minor changes, but otherwise good to go!

PR #762, opened 2020-07-17
Simplify dockerfile to node 10 image

sruti:
Simplifies the dockerfile
As a bonus, Node 10 has an upgraded OpenSSL version so it should solve SSL cert security errors like:
```
{ Error: certificate has expired
    at TLSSocket.<anonymous> (_tls_wrap.js:1116:38)
    at emitNone (events.js:106:13)
    at TLSSocket.emit (events.js:208:7)
    at TLSSocket._finishInit (_tls_wrap.js:643:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:473:38) code: 'CERT_HAS_EXPIRED' }
undefined



Couldn't download the latest Root CAs, but it's not a big deal.

Use "require('ssl-root-cas')" instead of "require('ssl-root-cas/latest')"
```


Dockerfile
```diff
@@ -55,10 +21,4 @@ COPY sources /app/sources/
...
 COPY migrations /app/migrations/
 
-#############################
-# entrypoint
-#
-RUN source $NVM_DIR/nvm.sh
-ADD .build_scripts/entrypoint.sh /
-RUN chmod +x /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
+CMD ["npm", "start"]
```
dolugen, line 24: (this might break the ECS task definition, please take this as a suggestion)

Perhaps we can set `npm` as the ENTRYPOINT, and set `start` as the default command like so:

```
ENTRYPOINT ["/usr/local/bin/npm"]
CMD ["start"]
```

This way people can only run `npm` commands, but can override the default command if they wish.

For example, this will run the default `start` command:
```
docker run --rm openaq_fetch:latest
```

While these will override the `start` command:
```
docker run --rm openaq_fetch:latest test 
docker run --rm openaq_fetch:latest start -- --dryrun -s "Beijing US Embassy"
```

sruti: I really like this suggestion but I couldn't get it to work, I might have just been doing something wrong. I'll definitely keep it in mind and hopefully we can implement it in future changes!

Issue #776, opened 2020-12-08
New Data: Quito


sruti:
We have a new data source from the government of Quito, with an API!

- [X]  Data is ‘raw’ and reported in physical concentrations not in an 'Air Quality Index' or equivalent (e.g. AQI, PSI, API) format, with units (ug/m3 or ppm)

- [X] Data is at the ‘station-level' with geographic coordinates, not aggregated into a higher (e.g. city) level (coordinates found by clicking on each location)

- [X] Data is from measurements averaged between 10 minutes and 24 hours, and we know whether the timestamp is for the beginning or end of the measurement period

- [X] Programmatic access!

API call to get all stations and metadata: http://www.aqvisor.net:9090/openaq/location

sruti:
# API Documentation
## for integration REMMAQ-OpenAQ

_Valeria Díaz, Juan Salvador, Paul Romero_

October 30th, 2020 - v 0.8

---

## `dataset`


This API will return a dataset where every record 
contains surveyed data for vector: 
**local time, magnitudes,** and **observation point**
in the format:

```
    {
        "date": {
            "local": "2020-10-21T04:00:00.000-05:00",
            "utc": "2020-10-21T04:00:00.000Z"
        },
        "parameter": "PM 10",
        "unit": "µg/m³",
        "value": 2.27346,
        "averagingPeriod": {
            "value": 24,
            "unit": "hours"
        },
        "location": "Carapungo",
        "city": "Quito",
        "country": "EC",
        "coordinates": {
            "latitude": -0.09548,
            "longitude": -78.44988
        },
        "sourceType": "government",
        "sourceName": "IMQ-AMBIENTE-REMMAQ",
        "mobile": "false",
        "attribution": [
            {
                "name": "REMMAQ",
                "url": "http://www.quitoambiente.gob.ec/ambiente"
            },
            {
                "name": "Secretaría del Ambiente"
            },
            {
                "name": "Municipio del Distrito Metropolitano de Quito"
            }
        ]
    }
```

#### Parameters

* **`magnitude`**

Is one of the parameters required by openaq.org: PM 2.5, PM 10, CO, NO2, SO2, O3. Please, note that **REMMAQ does not process BC**. 

If you don't specify this parameter, your dataset will return records for all 6 magnitudes.

If the parameter `magnitude` has a value different from the required, you will get
an error and message:

```
    {
[...]

PR #779, opened 2020-12-19
Add: Adapter/quito

vcc-LG:
Solves this issue: 
https://github.com/openaq/openaq-fetch/issues/776

Added adapter to Quito in Ecuador. 

_(This is my first contribution to the repo so if I've messed up any conventions or you want anything changed just let me know 👍 )_

### DIFF
--- a/adapters/quito.js
+++ b/adapters/quito.js
@@ -1,0 +1,135 @@
+'use strict';
+
+import { REQUEST_TIMEOUT } from '../lib/constants';
+import { default as baseRequest } from 'request';
+import { default as moment } from 'moment-timezone';
+
+const request = baseRequest.defaults({ timeout: REQUEST_TIMEOUT });
+
+var locations = [
+  'CML',
+  'COT',
+  'GUA',
+  'TUM',
+  'CH',
+  'P',
+  'ITC',
+  'CAR',
+  'JIP',
+  'BEL',
+  'CEN'
+];
+var measurementTypes = ['PM 2.5', 'PM 10', 'CO', 'NO2', 'SO2', 'O3'];
+
+const renameParameter = function (parameter) {
+  switch (parameter) {
+    case 'NO2':
+      return 'no2';
+    case 'SO2':
+    case 'Sulfur Dioxide':
+      return 'so2';
+    case 'CO':
+      return 'co';
+    case 'PM2.5':
+    case 'PM 2.5':
+      return 'pm25';
+    case 'O3':
+      return 'o3';
+    case 'PM10':
+    case 'PM 10':
+      return 'pm10';
+    default:
+      return parameter;
+  }
+};
+
+const sanitizeDate = function (date) {
+  try {
+    var utc = moment.tz(date['utc'], 'America/Guayaquil');
+    var local = moment.tz(date['local'], 'America/Guayaquil');
+    return {
+      utc: utc.toDate(),
+      local: local.format()
+    };
+  } catch (error) {
+    console.log('here');
+  }
+};
+
+const formatData = function (source, data) {
+  let measurements = [];
+  let dataObject = JSON.parse(data);
+  dataObject.forEach((element) => {
+    let m = {
+      location: element['location'],
+      value: element['value'],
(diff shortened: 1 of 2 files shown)


adapters/quito.js
```diff
@@ -0,0 +1,139 @@
...
+        sourceType: element['sourceType'],
+        sourceName: element['sourceName'],
+        mobile: element['mobile'] === 'true'
+      };
+      allMeasurements.push(m);
+    });
+    return allMeasurements;
+  }
+
+  locations.forEach((station) => {
```
sruti, line 97: I would suggest using the endpoint `http://www.aqvisor.net:9090/openaq/location` to programmatically get all the stations (that way in the future if they add stations we won't have to change the adapter)

adapters/quito.js
```diff
@@ -0,0 +1,139 @@
...
+          try {
+            var data = formatData(source, body);
+
+            if (data === undefined) {
+              return cb({ message: 'Failure to parse data.' });
+            }
+          } catch (e) {
+            return cb({ message: 'Unknown adapter error.' });
+          }
+          return cb(null, {
```
sruti, line 131: Because of the way this is structured, it's actually only returning 1 measurement. We want to parse all the measurements and return them all at once. My suggestion to simplify things would be to use promises similar to the [Icelandic adapter](https://github.com/openaq/openaq-fetch/blob/develop/adapters/ust-ist.js).

sources/ec.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+      "url": "http://www.aqvisor.net:9090/openaq/dataset",
+      "adapter": "quito",
+      "name": "REMMAQ",
+      "city": "Quito",
+      "country": "EC",
+      "description": "",
+      "sourceURL": "http://www.aqvisor.net:9090/openaq/location",
```
sruti, line 9: sourceURL should be a link to the data source entity, so in this case: `http://www.quitoambiente.gob.ec/ambiente/`

sources/ec.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+      "url": "http://www.aqvisor.net:9090/openaq/dataset",
```
sruti, line 3: I might change this to just be `http://www.aqvisor.net:9090/openaq/` so that you can use the `location` and `dataset` endpoints in the adapter

adapters/quito.js
```diff
@@ -0,0 +1,139 @@
...
+  'CAR',
+  'JIP',
+  'BEL',
+  'CEN'
+];
+const measurementTypes = ['PM 2.5', 'PM 10', 'CO', 'NO2', 'SO2', 'O3'];
+
+const request = baseRequest.defaults({ timeout: REQUEST_TIMEOUT });
+
+const renameParameter = function (parameter) {
```
sruti, line 26: You may be able to use the util function [unifyParameters](https://github.com/openaq/openaq-fetch/blob/develop/lib/utils.js#L59) here

adapters/quito.js
```diff
@@ -0,0 +1,139 @@
...
+        unit: element['unit'],
+        parameter: renameParameter(element['parameter']),
+        averagingPeriod: element['averagingPeriod'],
+        date: sanitizeDate(element['date']),
+        coordinates: element['coordinates'],
+        attribution: element['attribution'],
+        city: element['city'],
+        country: element['country'],
+        sourceType: element['sourceType'],
+        sourceName: element['sourceName'],
```
sruti, line 89: `sourceName` should match what's in the source file, so the source file should be updated with whatever it's returning here.

sruti (changes requested):
It's almost there! It's reading through all the data but only returning 1 measurements so order of operations needs to be tweaked a bit. And it would be great to get the list of stations programmatically. I think the [Icelandic adapter](https://github.com/openaq/openaq-fetch/blob/develop/adapters/ust-ist.js) could be a good model for how to structure this. Added a few other small suggestions/improvements too. Thank you!

sruti:
Closing in favor of #785

PR #797, opened 2021-03-22
Modify fetch to stop writing to DB

sruti:
With the new DB, ingest happens directly from S3 into the DB via a lambda function. This means fetch no longer has to write to the DB.

Note: I left in the code to post to the webhook (although that doesn't do anything currently) because I'll be revisiting the webhook -> fetches table -> fetches API -> fluffy/status/etc flow in a future improvement

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -1,4 +1,3 @@
-import knex from 'knex';
 import moment from 'moment';
 import { pick } from 'lodash';
 import S3UploadStream from 's3-upload-stream';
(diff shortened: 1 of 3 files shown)


sruti:
If there was an easy way to get the # of measurements written to S3 (count # of lines?), that could nicely replace the count we were getting back from items inserted in the db. Although trying to count the # of lines might ruin the efficiency of streaming the data.

lib/db.js
```diff
@@ -1,115 +1,10 @@
...
-      }
-
-      counts.inserted++;
-      return true;
-    })
-  ;
-}
-
 function saveResultsToS3 (output, s3, bucketName, key, s3ChunkSize) {
   const stream = new DataStream();
```
jflasher, line 10: On https://github.com/openaq/openaq-fetch/pull/797/files#diff-69f9a48e7478eb5df50c35b61f5638a078f274a604fff85d69732c2651ef8b47R13, I am not sure `status` is defined anymore? This may cause some of the data to be stripped off here, unsure.

jflasher:
It's a bit hard for me to provide super useful feedback just looking over the code, but assuming it runs locally fine, I think this should be ok. I added a comment that may or may not be an issue. My biggest worry would be that something is missing that was causing saving to be kicked off. But assuming saving runs and the process waits to exit until saving is done, then I think it'd be ok.

sruti:
I tested it out to make sure it properly writes to S3 (you can see files under `test-realtime` in `openaq-fetches`) and the process finishes only after so it should be fine. I will also test it on staging just to be sure.

lib/db.js
sruti: The `status` here is only checking the duplication in the database and it doesn't show up anywhere else. But since we're not writing to the DB anymore, it doesn't matter

PR #803, opened 2021-04-10
Update dev-quick-start to match defaults in env.js for first time setup

dabutvin:
the defaults in [env.js](https://github.com/openaq/openaq-fetch/blob/develop/lib/env.js) looks like this:

```js
  const psqlPassword = _env.PSQL_PASSWORD || 'openaq-pass';
  const psqlDatabase = _env.PSQL_DATABASE || 'openaq-local';
```

`openaq-pass` vs `openaq-password`
and
`openaq-local` vs `openaq`

This was a snag I hit following the setup guide for the first time.
Once I got them aligned i was able to get everything running.


sruti:
Good catch! Thanks for updating it.

PR #825, opened 2021-08-23
Disable Bosnia2 adapter, because service is gone

Rub21:
@kamicut @yunica


Rub21:
Closing this PR , because the service can still working some hours and some not, since the adapter does not stop the process, we just leave as it tis.

PR #830, opened 2021-08-24
add three new stations from qualar.cetesb.sp.gov.br

yunica:
@Rub21


Rub21:
Thank @yunica , this looks good!

PR #848, opened 2021-09-06
activate mexico, update timezone and  switch

yunica:
@Rub21


Rub21:
Thanks @yunica , Looks good and works fine this changes!

PR #908, opened 2022-10-18
UAE fetcher

majesticio:
config and fetcher for UAE from adairquality.ae


russbiggs:
@majesticio change this to merge into ```master```

PR #911, opened 2022-10-21
Turkiye Fetch is born

majesticio:
Added adapters/turkiye.js and updated sources/tr.json config file

source url: https://sim.csb.gov.tr/Services/GetAirQualityStations?type=0

sim_site_screenshot


adapters/turkiye.js
```diff
@@ -0,0 +1,105 @@
...
+ 
+ /**
+  * Fetches the data for a given source and returns an appropriate object
+  * @param {object} source A valid source object
+  * @param {function} cb A callback of the form cb(err, data)
+  */
+  
+ exports.fetchData = function (source, cb) {
+ 
+     const ALL_DATA =
```
russbiggs, line 26: This should be coming from the source JSON. e.g. https://github.com/openaq/openaq-fetch/blob/master/adapters/airnow-http.js#L73

sources/tr.json
```diff
@@ -11,5 +11,18 @@
             "info@openaq.org"
         ],
         "active": true
+    },
+    {
+        "url": "https://sim.csb.gov.tr/SERVICES/airquality",
```
russbiggs, line 16: This should be what you have in sourceURL and sourceURL should be this

sources/tr.json
```diff
@@ -11,5 +11,18 @@
...
+        "adapter": "turkiye",
+        "name": "Turkey",
```
russbiggs, line 18: lets change name to "Turkiye"

russbiggs (changes requested):

adapters/turkiye.js
majesticio: so I can replace ALL_DATA with source.url?

russbiggs: yeahj

russbiggs (changes requested):

russbiggs (approved):

Issue #918, opened 2022-10-27
Australia - South Australia Fetch error


caparker:
The data being fetched by the adapter has changed and so the adapter needs to be updated. I added a few things to be more consistent with the error handling but did not fix the data issues. First issue is that the site names have changed. 

If we do continue to have site information hard coded into adapters than we should probably create an error that reflects that the underlying data has been modified. 

You can review the error by switching to the `lambda-deployment` branch  and then running
```bash
node index.js -e .env -ds 'Australia - South Australia'
```

majesticio:
Thanks for pointing that out I'm going to look into updating this is adapter. The added error handling should come in handy.

PR #957, opened 2023-02-28
Bug/south australia

majesticio:
Ready for review/PR. Wiki available at https://github.com/openaq/openaq-fetch/wiki/South-Australia-Environmental-Protection-Authority-(EPA)


majesticio:
closes #918

src/adapters/southaustralia.js
```diff
@@ -0,0 +1,233 @@
...
+                },
+                attribution: [
+                    {
+                    name: "South Australia Environmental Protection Authority (EPA)",
+                    url: "https://data.sa.gov.au/data/dataset/recent-air-quality" 
+                    }
+                ],
+                averagingPeriod: {
+                    unit: "hours",
+                    value: 1
```
russbiggs, line 129: This doesnt line up for CO, since its an 8 hour averaging period

russbiggs (changes requested):

src/adapters/southaustralia.js
russbiggs: each parameter is stored separately so it is possible. You'll likely need to make `value` a dynamic value based on some sort of `if..else...` logic and toggle between 8 and 1.

PR #966, opened 2023-04-03
added function to remove duplicate stations

majesticio:
removed measurement locations that are deemed to be duplicates of of EEA data. Made a note on the Catalonia wiki page.

### DIFF
--- a/src/adapters/catalonia.js
+++ b/src/adapters/catalonia.js
@@ -92,7 +92,7 @@ function formatData(data) {
   
   const allData = concatAll(Object.values(data.map(aqRepack)));
   let measurements = getLatestMeasurements(allData);  
-  measurements = filterDuplicates(measurements, denyList) // remove duplicates of EEA stations!
+  measurements = filterDuplicates(measurements, eeaCataloniaDuplicateStations) // remove duplicates of EEA stations!
   return { name: 'unused', measurements: measurements };
 
 }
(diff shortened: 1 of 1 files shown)


src/adapters/catalonia.js
```diff
@@ -107,4 +107,95 @@ const getLatestMeasurements = function (measurements) {
...
+
+      return matchLocation && matchCity && matchCoordinates && matchParameter;
+    });
+  });
+}
+
+const blacklist = [
+  {
+      "location": "Alcover",
+      "city": "Alcover",
```
russbiggs, line 130: It looks like the location and the city fields are always the same so a bit redundant we could just remove one

src/adapters/catalonia.js
```diff
@@ -107,4 +107,95 @@ const getLatestMeasurements = function (measurements) {
     }
   });
   return Object.values(latestMeasurements);
-}
\ No newline at end of file
+}
+// remove stations deemed to be duplicates of EEA stations
+function filterDuplicates(measurements, criteria) {
+  return measurements.filter(measurement => {
```
russbiggs, line 113: are we checking each parameter here to see if it is duplicated? seems like it would be best to just wholesale exclude an entire station if anything is reported. The provider gets linked at the sensor_node level not at the sensor level. i.e. we cant get PM2.5 from provider A and PM10 from provider B and be able to show that distinction

russbiggs (changes requested):

src/adapters/catalonia.js
russbiggs: right but youre checking at the measurements level using the `measurement.parameter === criterion` seems like thats an unecessary check if we can just check at the station level

majesticio: ok done. just to be clear going for I should only check for duplicates using coordinates, not coordinates + parameters?

russbiggs (approved):

PR #968, opened 2023-04-04
update to averaging period

majesticio:
corrected to `unit: "minutes"`


majesticio:
`averagingPeriod: {unit: 'minutes', value: 10} `

PR #986, opened 2023-04-11
updated according to rubric

majesticio:
adapter now uses parseFloat, let, luxon and got

### DIFF
--- a/src/adapters/au_act.js
+++ b/src/adapters/au_act.js
@@ -1,32 +1,29 @@
 'use strict';
 
 import { REQUEST_TIMEOUT } from '../lib/constants.js';
-import { default as baseRequest } from 'request';
 import cloneDeep from 'lodash/cloneDeep.js';
 import flatten from 'lodash/flatten.js';
-import { default as moment } from 'moment-timezone';
-const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+import { DateTime } from 'luxon'
+
+import got from 'got';
 
 export const name = 'au_act';
 
 export function fetchData (source, cb) {
-  // get the time 1 day ago in AEST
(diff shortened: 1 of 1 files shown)


src/adapters/au_act.js
```diff
@@ -1,66 +1,71 @@
...
-      return cb({message: 'Failure to load data url.'});
-    }
-    // Wrap everything in a try/catch in case something goes wrong
+  }).then(res => {
+    const body = res.body;
+    
     try {
-      // Format the data
-      var data = formatData(JSON.parse(body), source);
+      let data = formatData(JSON.parse(body), source);
```
russbiggs, line 26: this can be `const` its never changed

src/adapters/au_act.js
```diff
@@ -1,66 +1,71 @@
...
+  let types = {
     'no2': 'no2',
     'o3_1hr': 'o3',
     'co': 'co',
     'pm10': 'pm10',
     'pm2_5': 'pm25'
   };
 
-  var units = {
+  let units = {
```
russbiggs, line 60: this should be `const` its never reassigned

src/adapters/au_act.js
```diff
@@ -1,66 +1,71 @@
...
+  
+}
 
-var formatData = function (data, source) {
-  var parseDate = function (string) {
-    var date = moment.tz(string, 'Australia/Sydney');
-    return {utc: date.toDate(), local: date.format()};
+
+let formatData = function (data, source) {
+  let parseDate = function (string) {
```
russbiggs, line 43: this should be `const` its never reassigned

russbiggs (changes requested):

majesticio:
merged into update-libraries

PR #988, opened 2023-04-11
updated EPA Victoria adapter according to rubric

majesticio:


majesticio:
merged into update-libraries

PR #995, opened 2023-04-13
replaced deprecated libraries for buenos-aires.js

majesticio:
removed and replaced deprecated libraries.
- `moment` -> `luxon`
- `request` -> `got`


majesticio:
merged into update-libraries

PR #997, opened 2023-04-13
replaced deprecated libraries for chile.js

majesticio:
- `moment` -> `luxon`
- 'request` -> `got`
I added retries to make sure that the url will resolve, as it seemed inconsistent.


majesticio:
merged into update-libraries

PR #998, opened 2023-04-14
replaced deprecated libraries for defra.js

majesticio:
- `moment` -> `luxon`
- `request` -> `got`


majesticio:
merged into update-libraries

PR #1002, opened 2023-04-18
new ACUMAR adapter ready for review/PR

majesticio:


src/adapters/acumar.js
```diff
@@ -0,0 +1,136 @@
...
+
+export async function fetchData(source, cb) {
+  try {
+    const results = await Promise.all(
+      stations.map((station) => getPollutionData(station))
+    );
+
+    const flattenedResults = results.flat();
+
+    // console.dir(flattenedResults, {
```
russbiggs, line 37: remove comment

src/adapters/acumar.js
```diff
@@ -0,0 +1,136 @@
+/**
+ * This code is responsible for implementing all methods related to fetching
+ * and returning data for the ACUMAR Argentina data source.
+ */
+
+'use strict';
+
+import got from 'got';
```
russbiggs, line 8: `got` is missing from package.json in this branch

src/sources/ar.json
```diff
@@ -13,5 +13,20 @@
             "info@openaq.org"
         ],
         "active": true
+    },
+     {
+        "url": "http://jmb.acumar.gov.ar/calidad/contaminantes${slug}.php",
```
russbiggs, line 18: instead of using the slug, because you can't pass is as a template string with the variable when its used as a simple string here lets just showthe `http://jmb.acumar.gov.ar/calidad/contaminantes.php` here since that is kind of "landing page" even though its also technically a station specific page

russbiggs (changes requested):

russbiggs:
Ive noticed the PM2.5 and PM10 have a bit of lag in their reporting sometimes, we may want to think about a strategy to handle this:

Screenshot 2023-04-24 at 3 57 05 PM

russbiggs:
There appears to be some bugs on the site where the date column does not consistently show the correct date. e.g. this following screenshot is from today 25/4/2023 but all the dates show 23/4/2023 
Screenshot 2023-04-25 at 10 52 58 AM

majesticio:
I created an offset that is keyed off of source.datetime. Please double check this for me to make sure we are on the same page @russbiggs

PR #1004, opened 2023-04-20
replaced deprecated libraries laqn.js

majesticio:
updated:
- `moment` -> `luxon`
- `request` -> `got`
I need to be sure the format of `source.datetime` before we pull this one in... would it be the format of `datetime` referenced in [index.js](https://github.com/openaq/openaq-fetch/blob/6c09adbbe0ec695e8a3f924a0645ee65af0b64cf/src/index.js#L11)?

### DIFF
--- a/src/adapters/laqn.js
+++ b/src/adapters/laqn.js
@@ -1,17 +1,16 @@
 'use strict';
 
 import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


majesticio:
merged into update-libraries

PR #1005, opened 2023-04-20
replaced deprecated libraries for medellin.js

majesticio:
replaced deprecated libraries for medellin.js
- `moment` -> `luxon`
- `request` -> `got`
I also corrected a datetime issue that I discovered, it was using the utc datetime as the local datetime.


majesticio:
merged into update-libraries

PR #1010, opened 2023-04-25
removed unneccesary packages, added needed ones

majesticio:
cleaned up dependencies. 'depcheck' is listed but we need that to check the depends


russbiggs:
`depcheck` doesn't ignore itself? if that really is the case lets add some additional command to the command to keep it less noisy. 

also looks like `winston-papertrail` is included but unused

PR #1011, opened 2023-04-27
Khartoum Residential active: false. Duplicate of airnow

### LINKED ISSUE #1009: Duplicate sources dosairnow vs airnow
It seems like some of the Department of State (DoS) locations may be duplicate fetching from the DoS and the Airnow-http sources. e.g. 

https://github.com/openaq/openaq-fetch/blob/main/src/sources/sd.json#L18

This shows the `Khartoum Residential` location is coming from the [dosairnowdata](https://www.dosairnowdata.org/) source but on the explore it show simply as "Airnow"

@majesticio can you look into this and see if this data being fetched from both sources? You'll likely need to check the original sources and check our s3 bucket to see if its landing there from both sources. if so we'll want to turn off one, likely the dosairnow so we arent pulling the same data from two places.

majesticio:
Every measurement of this adapter is the same location, and a duplicate of AirNow data. Setting to active: false
closes #1009

### DIFF
--- a/src/sources/sd.json
+++ b/src/sources/sd.json
@@ -23,6 +23,6 @@
         "contacts": [
             "info@openaq.org"
         ],
-        "active": true
+        "active": false
     }
 ]


russbiggs (commented):
Can you add this change to the changelog as well with an explanation for why we're removing it

majesticio:
merged into update-libraries

PR #1017, opened 2023-05-02
updated fetch.js to use luxon instead of moment

majesticio:
`moment` -> `luxon`


majesticio:
merged into update-libraries

PR #1029, opened 2023-05-18
updated deprecated library for trinidadtobago.js

majesticio:
updated `request` -> `got`. added `tough-cookie` library as `got` does not have a built in process for this.


majesticio:
merged into update-libraries

PR #1033, opened 2023-06-01
New data from Basilicata, Italy

majesticio:
~68 Stations from Basilicata. See [wiki](https://github.com/openaq/openaq-fetch/wiki/ARPA-Basilicata-(Italy)) for details.
Adapter takes 10-20 seconds to run.

### DIFF
--- a/src/adapters/basilicata.js
+++ b/src/adapters/basilicata.js
@@ -164,7 +164,7 @@ async function getBearerAuth() {
 }
 
 let translations = {
-  // other values are from the same locations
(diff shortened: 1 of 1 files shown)


russbiggs:
We'll want to table this, because we do already get this data from EEA, with the exception of the 24 hour PM2.5 data. Adding this would cause conflicting sensor_nodes with the existing sites

majesticio:
This adapter has been determined to be duplicate EEA data. This adapter can be updated and merged if the need arises.

PR #1042, opened 2023-07-14
ACUMAR - offset

majesticio:
Updated the sourceName to 'acumar-ar'. Having the sourceName the same as the adapter name was causing issues with data ingestion.


russbiggs:
@majesticio i think this PR is no longer relevant, is that correct? if so lets close

majesticio:
this is relevant, and has necessary corrections

PR #1044, opened 2023-07-14
Rio fix - updated to use got instead of fetch

majesticio:


majesticio:
I believe using 'got' will make this adapter more robust. The url for this source is finicky and got's retry feature should help the url resolve if it doesn't work the first time its called

majesticio:
closing. Adapter for Rio has been updated to use our got client

PR #1045, opened 2023-07-14
removed files where active:false & updated changelog

majesticio:


russbiggs:
can we move the CHANGELOG.md to the project root, and it seems like we now have the unused `transliteration` package, can you remove that for this PR as well.

majesticio:
Ok, the requested changes have been made

PR #1046, opened 2023-07-17
Updated Spartan adapter to only fetch recent records

majesticio:
Significantly decreases payload size, should now work nicely

### DIFF
--- a/src/adapters/spartan.js
+++ b/src/adapters/spartan.js
@@ -12,44 +12,6 @@ import got from 'got';
 import { parse } from 'csv-parse';
 import { DateTime } from 'luxon';
 
-const slugs = [ // commented out ones are missing data, might be useful later
-  'AEAZ',
-  'ARCB',
-  'AUMN',
-  'BDDU',
-  // 'BIBU',
-  'CADO',
-  'CAHA',
-  'CAKE',
-  'CALE',
-  'CASH',
-  'CHTS',
-  // 'CLST',
-  // 'CODC',
-  // 'ETAD',
-  'IDBD',
-  // 'ILHA',
-  'ILNZ',
-  // 'INDH',
-  'INKA',
-  'KRSE',
-  'KRUL',
-  'MXMC',
-  'NGIL',
-  // 'PHMO',
-  // 'PRFJ',
-  'SGSU',
-  // 'TWKA',
-  // 'TWTA',
-  // 'USBA',
-  // 'USBO',
-  // 'USMC',
-  // 'USPA',
-  'VNHN',
-  // 'ZAJB',
-  'ZAPR',
-];
-
 export const name = 'spartan';
 
 export async function fetchData (source, cb) {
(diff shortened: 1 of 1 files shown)


russbiggs:
Can you add a wiki entry for SPARTAN that helps describe the folder structure etc so that I can review this with more context?

majesticio:
[spartan wiki](https://github.com/openaq/openaq-fetch/wiki/Spartan-Network)

russbiggs:
So one issue i see if we're not using the daily files that they provided for us e.g.: 

http://data.spartan-network.org/GroupedBySite/CHTS/TimeRes_ByDate/

These files would allow more focused searching and not require loading all the data into memory, especially in the case of really large files e.g. http://data.spartan-network.org/GroupedBySite/CASH/TimeResPM25_HourlyEstPM25_CASH.csv which already weighs in at almost 10MB. 

The irregular update schedule of the data does seem really problematic in that with either method i'm not sure how we can reliably stay up-to-date without grossly over-fetching. I'm wondering if this would benefit from a different approach more akin to a bulk load that we do monthly?

majesticio:
Most all of the files in the directory linked have not been updated in a long time. I had to comb through all of the available folders to find the ones that are kept up to date. I don't think RAM was our issue when the adapter was erroring out. Another idea would be to run it once a day for the last 24hr of measurements. I think if we do a huge one once a month we might run into the same issue we had with the adapter error again

majesticio:
I also think this PR will get it to run properly in its current state

russbiggs:
The problem is as these file grow the use of RAM will continue to go up and in turn the amount of time it takes to process will also increase. Crunching 10MB files every hour or every day will be more expensive on the lambda, this will not be a sustainable method. I specifically requested the daily files so we should be using those if possible. If they aren't being updated at a good interval we can communicate with the SPARTAN team.

majesticio:
Fair point. I suppose we should reach out then, as the current method will become untenable over time

russbiggs:
even in the full files stuff isnt getting updated very recently, heres all the sites by their most recent measurement date: 
site_code | date
-- | --
AUMN | 2023-06-19T17:00:00
USNO | 2023-06-17T23:00:00
CASH | 2023-06-16T10:00:00
AEAZ | 2023-06-07T17:00:00
MXMC | 2023-05-31T11:00:00
KRUL | 2023-01-06T13:00:00
CHTS | 2022-06-12T20:00:00
CAHA | 2022-04-06T14:00:00
ZAPR | 2021-12-04T13:00:00
KRSE | 2021-03-16T03:00:00
NGIL | 2020-10-13T02:00:00
VNHN | 2020-06-23T14:00:00
SGSU | 2020-01-28T18:00:00
INKA | 2019-11-25T13:00:00
CAKE | 2019-11-14T09:00:00
CADO | 2019-09-05T11:00:00
CALE | 2019-05-14T15:00:00
IDBD | 2018-01-29T10:00:00
ARCB | 2017-01-23T15:00:00
BDDU | 2015-09-21T06:00:00

PR #1049, opened 2023-07-18
update ACUMAR adapter to fetch more rows for ingestion

majesticio:


russbiggs:
this looks good, can you add a deployment to deployment.js with maybe a 12 hour offset?

russbiggs:
im sorry we discussed doing a 72 hour offset didn't we? I think thatll be better given the update frequency issues.

majesticio:
Should be good to go

PR #1051, opened 2023-07-21
Acumar offset update

majesticio:
if source.datetime is true it will use the offset to find the appropriate row


russbiggs:
@majesticio can you fix the conflict?

majesticio:
I got the conflicts resolved

PR #1053, opened 2023-07-25
Fix/canterbury data

majesticio:
Fixed problems that prevented the adapter from fetching data from all available stations.


russbiggs:
A lot going on here, can you give a synopsis of what changed?

majesticio:
The adapter was not picking up all the stations. The basic idea of making a list of tasks and running them in parallel is the same, however I was not able to troubleshoot the original logic, so I rewrote a lot of it. I also added a dictionary for all of the source's parameter keys that is more explicit.

PR #1055, opened 2023-08-01
fixed datetime errors for tasmania

majesticio:
timezone was being incorrectly assigned.

### DIFF
--- a/src/adapters/tasmania.js
+++ b/src/adapters/tasmania.js
@@ -24,7 +24,7 @@ export function fetchData (source, cb) {
         if (data === undefined) {
           throw new Error('Failure to parse data.');
         }
-
+console.dir(data, {depth: null})
         cb(null, data);
       } catch (error) {
         cb(error, { message: 'Unknown adapter error.' }, null);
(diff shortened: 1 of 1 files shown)


russbiggs:
can you provide a bit more detail on how the timezone was incorrect? do we need to correct all the previous data?

majesticio:
original:
```
const date = DateTime.fromFormat(
      string.trim(),
      'HHmmss',
      'Australia/Hobart'
    );
```    
update:
```
const date = DateTime.fromFormat(
      string.trim(),
      'HHmmss',
      { zone: 'Australia/Hobart' }
    );
```
old measurement obj:
```
 {
      location: 'Cygnet',
      city: 'Tasmania Region',
      date: {
        utc: '2023-08-01T11:10:07Z',
        local: '2023-08-01T05:10:07-06:00'
      },
    }
```

new measurement obj:
```
{
      location: 'Cygnet',
      city: 'Tasmania Region',
      date: {
        utc: '2023-08-01T19:10:07Z',
        local: '2023-08-02T05:10:07+10:00'
      }
}
```      

Unfortunately it looks like the fetch data will need to be updated.

russbiggs:
@majesticio OK so it looks like the data we have is off due to the timezone issue, but why does this cause a 15 hour gap in the measurements?

majesticio:
@russbiggs I can't understand why but `date.zone.name` returns `America/Denver` (UTC -6). With the fix it returns `Australia/Hobart` (UTC +10/11 ) so that seems to be the cause

russbiggs:
@majesticio That sounds like its pulling from your machine. I guess I'm curious how this fixes the gap though, there seems to be two disparate issues. Can you confirm (through testing) that this will fetch the values we are currently missing in that 15 hour gap? I just dont see how the timezone fix would resolve that, but I am perhaps not seeing some of the other logic of the adapter.

majesticio:
@russbiggs The AWS us-east-1 region operates in the Eastern Standard Time (EST, UTC -5). On the other hand, Australia/Hobart operates at Australian Eastern Standard Time (AEST, UTC +10). When I ran it locally with the unfixed code the adapter used my local machine time of Denver (UTC -6). An AWS instance running in Eastern Standard Time (EST, UTC -5) explains the 15 hour gap we see with the measurements.

PR #1062, opened 2023-10-05
luxon timezone correction for tasmania

majesticio:


majesticio:
Improved datatime object instantiation has been implemented, hopefully will fix the time discrepancy issue

PR #1070, opened 2024-01-09
Update Node version to 18

majesticio:
I have tested running all of the adapters several times after updating fromNode 14 to Node 18. I think we can deploy this safely, I will be sure to keep an eye on it.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,7 +31,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: "14"
+          node-version: "18"
 
       - name: Install CDK
         run: |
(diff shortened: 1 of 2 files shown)


russbiggs:
Sorry i had missed this getting assigned to me. The current changes only update the node environment for the action (which we currently dont use). Changes need to be made to the Lambda runtimes in the CDK stacks

majesticio:
Thanks for catching that. I updated the CDK to reflect that and use Node 18

majesticio:
Got it updated, thank you

majesticio:
> it appears this deployment worked without issues, if thats really the case i would go ahead and merge
Sounds good, I will get this merged and deployed.

Issue #1068, opened 2023-11-21
EEA using deprecated battuta service


russbiggs:
The EEA adapter is still using the deprecated Battuta service. It would be best to remove these calls to the Battuta file to lessen the load on the bucket:

https://github.com/openaq/openaq-fetch/blob/93496c79adb4e5038508f4e387ed75e469c50b43/src/adapters/eea-direct.js#L43

russbiggs:
@majesticio  Can this issue be closed

majesticio:
Closing this issue, was resolved by PR #1071

PR #1071, opened 2024-01-19
Removed battutaStream usage from eea-direct adapter

majesticio:
Removed battutaStream and fetchMetadata functions from eea-direct. Also added some conversions for incorrect units. Not relying on the battuta bucket may have added some locations to our EEA dataset as well.


majesticio:
closes #1068

russbiggs (commented):
Only comment is to resolve the now used packages from the unused packages bot `JSONStream`, `node-fetch`, `tz-lookup`

PR #1072, opened 2024-01-22
Add Taiwan EPA data

majesticio:
Rebased onto main, and ready to merge. 
Link to [Taiwan wiki](https://github.com/openaq/openaq-fetch/wiki/Taiwan-Environmental-Protection-Administration)

### DIFF
--- a/src/adapters/taiwan.js
+++ b/src/adapters/taiwan.js
@@ -1,0 +1,133 @@
+import got from 'got';
+import { DateTime } from 'luxon';
+import fs from 'fs/promises';
+
+const baseUrl =
+  'https://airtw.moenv.gov.tw/json/airlist/airlist_{i}_{dateStr}.json';
+const start = 1;
+const end = 150;
+const stationDataPath = './stations2.json';
+
+function createUrls(baseUrl, start, end, dateStr) {
+  return Array.from({ length: end - start + 1 }, (_, i) =>
+    baseUrl
+      .replace('{i}', (start + i).toString())
+      .replace('{dateStr}', dateStr)
+  );
+}
+
+async function fetchUrl (url) {
+  try {
+    return await got(url, {
+      headers: { Accept: 'application/json' },
+    }).json();
+  } catch (error) {
+    console.error(
+      `Error fetching or parsing data from: ${url}`,
+      error.response?.body
+    );
+    return null;
+  }
+}
+
+function combineData (airQualityData, stationData) {
+  return Object.entries(airQualityData).reduce(
+    (combined, [key, airQualityEntry]) => {
+      const stationEntry = stationData.find(
+        (station) => station.SiteName === airQualityEntry.sitename
+      );
+      if (
+        stationEntry &&
+        stationEntry.TWD97_Lon &&
+        stationEntry.TWD97_Lat &&
+        airQualityEntry.date
+      ) {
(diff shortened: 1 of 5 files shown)


src/adapters/taiwan.js
```diff
@@ -0,0 +1,163 @@
...
+
+'use strict';
+
+import got from 'got';
+import { DateTime } from 'luxon';
+
+import log from '../lib/logger.js';
+
+const start = 1;
+const end = 150;
```
russbiggs, line 14: not clear what these two values are doing

majesticio: This is a range of station IDs. It tries to find all stations with any ID from 1 to 150. This is how I originally found all the stations, but perhaps I should take a look at a more elegant solution

src/sources/xk.json
```diff
@@ -23,6 +23,6 @@
         "contacts": [
             "info@openaq.org"
         ],
-        "active": true
+        "active": false
```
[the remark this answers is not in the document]
majesticio: It should not be here, I am going to roll that back. I remember that Kosovo is timing out, I will address this with another branch / PR

src/adapters/taiwan.js
```diff
@@ -0,0 +1,163 @@
...
+
+export const name = 'taiwan';
+
+export async function fetchData (source, cb) {
+  const dateTimeOneHourAgo = DateTime.now()
```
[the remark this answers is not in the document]
majesticio: The measurements have to be requested with the date and time in the url, so if for instance they have 4pm measurements it is 4:45 pm you will miss the data. Essentially it sends back measurements after the time you attach to the url

src/adapters/taiwan.js
russbiggs: ah gotcha, im guessing there is not a dynamic source of location IDs available? in lieu of a comment describing this in the code a better variable name would probably suffice then something like could get it into a single variable:

```js
const locationIds = [...Array.from(new Array(150), (x, i) => i + 1)]
```

majesticio: ok I like that

src/sources/tw.json
```diff
@@ -12,5 +12,19 @@
...
         "active": false,
         "notes": "Updated API. Working on updating the adapter"
+    },
+    {
+        "url": "https://airtw.moenv.gov.tw/json/airlist/airlist_{i}_{dateStr}.json",
+        "adapter": "taiwan",
+        "name": "taiwan-moenv",
+        "city": "",
+        "country": "TW",
+        "description": "Taiwanese data from EPA",
```
russbiggs, line 22: change this to not use a demonym for the data. Something like Data from the Tawain "moenv" (Ministry of Environment?)

majesticio: That is reasonable

caparker:
@majesticio Did we mean to merge this before I had a chance to look at it? I want to make sure its not going to create duplicate locations once we deploy it.

majesticio:
> @majesticio Did we mean to merge this before I had a chance to look at it? I want to make sure its not going to create duplicate locations once we deploy it.

Apologies, I merged it but did not deploy. Reopened as #1075

caparker:
No worries and thanks for reopening, I think that makes it easier in case we need to change anything. I will get this looked at by end of day on monday.

PR #1075, opened 2024-01-26
Add Taiwan

majesticio:
Reopening this PR, was merged prematurely.


majesticio:
Citing comment from @russbiggs:
@caparker is also going to look at this to make sure that we can successfully conflate these sites with the old sites from the previous Taiwan provider on ingest

caparker:
Ok. Everything looks good on my end. I updated one part to make sure that the `source_name` for any new locations would match the `source_name` for the old locations. As of right now we are first using the spatial method to match ingest data to a sensor node and then falling back on the `source_name/site_name` method. 

An interesting note, I was using google translate to get some of the location names to see if they still match what we have in the db and I noticed that the stations are of at least 3 different types
- general
- industrial
- mobile

So that brings up a question regarding mobile locations. Do we consider it mobile only if its actively moving?  I think that makes sense the way that we are currently treating mobile however maybe we need to think about adding some station type data, when provided.

russbiggs:
How many of the locations are mobile? Does our ingestion process for fetch support mobile? I'm guessing we only really get mobile in the Lcs streams.

majesticio:
Thats interesting @caparker I had not noticed the mobile reference, I will investigate further. And no, I don't believe `openaq-fetch` has a mechanism to support mobile stations at the moment.

caparker:
I believe if the data has a `mobile` field when its ingested we will honor it, even in the non-lcs data. 

But I bet these arent moving around much. I assume they are more like the trailer style systems that can get towed from one spot to another. So when its moved it will have new lat/lng and therefor be made into a new sensor_node, which is what we would want.

russbiggs:
@majesticio what is the status on this?

majesticio:
It sounds like everything is ready to go, I am going to merge and deploy this adapter

PR #1080, opened 2024-02-08
cleaned up env.js

majesticio:
cleaned up env.js and removed knexfile Dockerfile and associated docs


majesticio:
I am going to remove the unused dependency (knex)

src/lib/env.js
```diff
@@ -9,13 +9,6 @@ const yargs = _yargs(hideBin(process.argv));
...
  * @param {number} logLevel
  * @param {boolean} logColor
- * @param {string} psqlHost
- * @param {number} psqlPort
- * @param {string} psqlUser
- * @param {string} psqlPassword
- * @param {string} psqlDatabase
- * @param {number} psqlPoolMin
- * @param {number} psqlPoolMax
  * @param {string} apiURL
```
russbiggs, line 12: I dont believe fetch communicates with the API any more, this and any functionality that is using this is likely stale, re-review this usage

src/lib/env.js
```diff
@@ -9,13 +9,6 @@ const yargs = _yargs(hideBin(process.argv));
...
  * @param {string} webhookKey
```
russbiggs, line 13: Are we using the webhook still? I dont believe so. This an its usage needs to be reviewed and removed if needed

src/lib/env.js
```diff
@@ -181,13 +174,6 @@ export default () => {
   const maxParallelAdapters = +_env.MAX_PARALLEL_ADAPTERS || 1024;
```
russbiggs, line 174: Is this still relevant now thats its on lambda? seems like something from the container version

russbiggs (commented):
Still some questionable variables in the code. Also the documentation in the README.md need to be updated to reflect all the changes here: https://github.com/openaq/openaq-fetch/blob/main/README.md#installing--running 

and  here

https://github.com/openaq/openaq-fetch/blob/main/docs/env.md

I'm also curious how `index.adapter.sh`, `index.sh`, `run-script.js`, `docker-compose.yml` are used if at all. If not used they need to be removed as well.

Is anything is `data_scripts` being run with the new process? they seem like most legacy adapters.

src/lib/env.js
majesticio: maxParallelAdapters is being used in fetch.js and is included in our .env files

src/lib/env.js
majesticio: webhook key is being called by fetch.js, which in turn is being called by index.js

src/lib/env.js
majesticio: API_URL is used in deploy.yml for GH secrets

majesticio:
I have removed all of the unused files, I need to investigate the use of API_URL in deploy.yml to fetch GH secrets

russbiggs:
> I have removed all of the unused files, I need to investigate the use of API_URL in deploy.yml to fetch GH secrets

Ignore anything in github action. All this clean up is agnostic of whats happning in Github Actions

src/lib/env.js
caparker: Hard to say - the impression that i get is that it will default to 2x the cores based on the documentation for the python version. It was set to 1 in production though.

caparker:
@majesticio Where are we standing on this branch?

majesticio:
This should be good to go, but I will need to adjust it to accommodate our recent changes

russbiggs:
My comments were never addressed re:API_URL and WEBHOOK_KEY These are not actively used and need to be cleaned up

majesticio:
I will get those taken care of.

src/lib/notification.js
```diff
@@ -35,36 +23,22 @@ async function publish(message, subject) {
...
-            log.debug(r);
-        });
-        log.info(`Dry run finished with ${successes.length} successes and ${failures.length} failures in ${(fetchReport.timeEnded - fetchReport.timeStarted)/1000} seconds`);
-
-        if (!env.dryrun) {
-		        await publish(fetchReport.results, 'fetcher/success');
-        }
-
-        return 0;
-    };
```
caparker, line 69: This all needs to stay

caparker (changes requested):
Overall it looks good. Just need to fix that function in the notifications.js file that I commented.

caparker (commented):
This one does not seem ready. I ran it locally and it did not finish and because there are so many changes it makes me nervous to merge it into main. 

Could you (@majesticio) run it locally and compare it to a local run on the main branch and let me konw what you think?

majesticio:
@caparker it ran identically for me, and finished a few seconds faster. I am concerned if it does not work on your end

caparker:
I’ll take a closer look on my end. May have just been a fluke .

PR #1081, opened 2024-02-09
Fixed parameters' units and updated adapter to overfetch to fill data gaps for Taiwan

majesticio:
To be merged and deployed after faulty new measurements and rollups have been deleted.

### DIFF
--- a/src/adapters/taiwan.js
+++ b/src/adapters/taiwan.js
@@ -6,14 +6,16 @@
 import log from '../lib/logger.js';
 import got from 'got';
 import { DateTime } from 'luxon';
+import {
+  unifyMeasurementUnits,
+} from '../lib/utils.js';
 
 export const name = 'taiwan';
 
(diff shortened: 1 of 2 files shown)


russbiggs:
Based on the data gaps i was seeing, i'm not clear how an offset deployment would help. What is the behavior we see that necessitates an additional deployment? Can the same be achieved with overfetching in the main deployment?

We would want to have a clear case for an additional deployment, outlining what the pattern is and why the gaps appear.

majesticio:
I have looked at it some more and I think we can try to overfetch instead of using an extra deployment as the measurements endpoint is queryable by the hour historically, I can update the adapter to fetch the last 3 or 4 hours of measurments. this should resolve the data gaps issue.

majesticio:
I have updated the adapter to overfetch by 3 hours to address the data gaps. I removed the deployment as we are now relying on the over fetching strategy

PR #1085, opened 2024-02-21
Peru OEFA adapter

majesticio:
Peru OEFA fetch adapter. Added custom error if an incorrect user / or password is used, and some debug statements for empty responses. This is because all responses from the API return `status 200` even if the wrong credentials are used. Waiting on source to confirm the dates are accurate before deploying as the measurements are all ~12+ hours old.


caparker (changes requested):
Could you 
- update this to use our request client
- add docstrings
-  change the parameters const at the top to follow the same pattern as what we are doing in the rwanda adapter. Use the object as a map to make sure we are getting the right units. 

And dont hesitate to reach out if any of that is unclear or you just have questions.

majesticio:
@caparker I added docstrings and will update the documentation. I was having a hard time getting the client to work with the custom errors we have, so I left `got` in place. Let me know what you think

caparker (approved):
Looks good

PR #1093, opened 2024-03-21
First draft of new eea adapter

caparker:
Not ready to be pulled into main but I wanted us to review it. 
Here is a list of things done in the branch
* New eea adapter with notes - missing docstrings I know :) 
* Added some parsing util functions (parseTimestamp, parseCoordinates)
* Added more return formats to the request
* Added a very early draft of a style guide. Also working on updating the dummy.js adapter (which we will rename)
* Updated the measurement schema, the eea data does not have a timezone in it. Plus we do not need it, timezone will come from coordinates.

@majesticio Please take a look and make sure everything in here is clear

### DIFF
--- a/src/adapters/README.md
+++ b/src/adapters/README.md
@@ -1,0 +1,33 @@
+# Adapter writing guidelines
+These are guidelines that should be followed whenever possible. If a specific one cannot be followed please document the reason in the code.
+
+## Purpose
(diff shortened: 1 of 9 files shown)


caparker:
Ok. Now we just need to make this adapter go back an hour or two and then do something about the source_names. Right now we are using `EEA xxxx` as the source name and we dont have that available in these data.

russbiggs:
Moving forward all EEA data should just be a single source. I've never liked the "EEA Italy" "EEA Belgium" since these are all just coming from a single source, EEA

Issue #1095, opened 2024-03-22
Ust-ist adapter failing


caparker:
@majesticio  could you look into this? I pinpointed the error and added comments inline so it should be easy to fix. 

Here is the branch
https://github.com/openaq/openaq-fetch/tree/adapter-fix/ust-ist

majesticio:
currently the data URL is returning
```
{
"status": "error",
"message": "no data"
}
```
so this adapter requires investigation

PR #1096, opened 2024-03-22
Adapter fix/ust ist

majesticio:
This adapter data URL `https://api.ust.is/aq/a/getLatest` does not appear to return data anymore. This has been replaced with `https://api.ust.is/aq/a/getDate/date/${yesterday}`. The URL seem to only work up till the day before the current date unfortunately but does return measurements. Endpoints can be investigated at`https://api.ust.is/aq/a`.

It has also been updated to over-fetch measurements to help with coverage.

### DIFF
--- a/cdk/stack.ts
+++ b/cdk/stack.ts
@@ -72,7 +72,7 @@ export class RealtimeFetcherStack extends cdk.Stack {
         description: `Fetcher for ${id}`,
         code: lambda.Code.fromAsset('../src'),
         handler: 'fetch.handler',
-        memorySize: 1024,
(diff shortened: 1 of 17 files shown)


majesticio:
closes #1095

caparker (approved):
Looks good. I removed a log line from the top of the file where you were logging out the value of yesterday. If that is something you wanted to leave in there then just add more to the log, like
```
log.debug('Showing value of yesterday', yesterday) 
```
Or something along those lines. Otherwise I will assume it was just for dev purposes and you meant to remove it.

PR #1097, opened 2024-03-27
Refactor au_vic.js and update source status in au.json

majesticio:
This adapter is rate limited by the source, takes ~30 seconds to run.

### DIFF
--- a/src/adapters/au_vic.js
+++ b/src/adapters/au_vic.js
@@ -1,15 +1,13 @@
 'use strict';
 
-import { REQUEST_TIMEOUT } from '../lib/constants.js';
-import { unifyMeasurementUnits } from '../lib/utils.js';
-
-import got from 'got';
 import _ from 'lodash';
 import { DateTime } from 'luxon';
-import parallelLimit from 'async/parallelLimit.js';
 import Bottleneck from 'bottleneck';
 
-const request = got.extend({ timeout: { request: REQUEST_TIMEOUT } });
+import { unifyMeasurementUnits } from '../lib/utils.js';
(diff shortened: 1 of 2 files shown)


caparker (changes requested):
It works but it could be cleaned/refactored

Lets follow the following rules:
* Keep things simple and readable
* declare global variables first and then functions
* fetchData should return the cb - I know that technically it is because you are passing it to formatData but refer to the first bullet, keep it simple. 
* formatData should only do one thing, take data and reshape it. Ideally I would like to see that formatData takes one row of data and reshapes that one row. And we dont need to use `unifyMeasurementUnits` that is done later in the `measurement.js` file
* Proofread your code once its done and passing all tests. Make sure that its not confusing, you have not left random `log.debug` lines.

majesticio:
Requested changes have been made for better readability and cleaner logging

caparker (changes requested):
Much better, but still some work to do. 

* Lets remove all use of lodash - we dont need it and deepclone is pretty heavy to use in this way.
* lets not be fetchingmeasurements in the formatdata function. It feels awkward.  How about we add something like fetchStationMeasurements, pass the stations to that and then loop and call fetchMeasurements
* Then you could loop through that response and for each measurement pass the measurement and station info to the formatData method. That way you dont need to use deepclone to prevent you from mutating an object all the time
* And if you still end up with nested arrays for some reason use [].flat() which is native.

caparker (approved):
Looks good. Most sites are not providing much data but there is nothing we can do about that. Seems like most sites only supply `particles`

PR #1098, opened 2024-03-28
Peru OEFA

majesticio:
Improved logging, and larger fetch window to improve coverage


caparker (approved):
Looks good. Once we have some hourly measurements we should go back and compare the data to the data on their map. Im just skeptical of those timestamps.

PR #1099, opened 2024-03-29
Update to `ust-ist` adapter for 'time beginning' measurements

majesticio:
Added 1 hour to compensate for 'time beginning' measurements

### DIFF
--- a/src/adapters/ust-ist.js
+++ b/src/adapters/ust-ist.js
@@ -12,12 +12,9 @@ import log from '../lib/logger.js';
 
 export const name = 'ust-ist';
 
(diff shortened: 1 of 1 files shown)


caparker (changes requested):
I hate to do this to you but it looks like the old datasource is back 
https://api.ust.is/aq/a/getLatest
Looks like its up to date and it goes back 24 hours so I think we should go back to using this source of data.

majesticio:
Good eye. It's better this way so I don't mind at all.

majesticio:
This adapter now uses the /latest endpoint for real time data.

closes #1095

caparker (approved):

PR #1103, opened 2024-04-09
Added check for credentials for the israel-envista adapter

majesticio:
This will throw an error if credentials or token is unavailable.


caparker (approved):
Looks good.

PR #1105, opened 2024-04-10
Update date sanitization logic in defra.js

majesticio:
A change to the date strings in HTML from this provider was causing this adapter to fail. This update uses regex to format the date string properly.


caparker (approved):
Thanks for removing the regex, I think this way is much simpler :)

PR #1106, opened 2024-04-11
Updated import and improved error handling for ARPALAZIO

majesticio:
This update improves some error handling in the `fetchData` function, and improves logging to make the debugging easier


majesticio:
Updated `fetchStream` which was not handling errors and passing unhandled promises occasionally

caparker (approved):
Looks good. I notice they are filtering out the errors and converting units so we will need to get a handle on which adapters are doing that but no reason to change it now.

PR #1107, opened 2024-04-12
recover new south wales

majesticio:
This is a draft. it appears to be returning valid measurements but is not finishing:
```
info: 0 new measurements found for "Australia - New South Wales" in 1.553s
info: Australia - New South Wales - 1 occurrences of Adapter error (au.json/nsw/'Australia - New South Wales'): Cannot read properties of undefined (reading 'readable')
```

### DIFF
--- a/src/adapters/nsw.js
+++ b/src/adapters/nsw.js
@@ -1,362 +1,130 @@
+/**
+ * This code is responsible for implementing all methods related to fetching
+ * and returning data for the New South Wales data sources.
+ */
+
 'use strict';
 
-import { REQUEST_TIMEOUT } from '../lib/constants.js';
-import { removeUnwantedParameters, convertUnits } from '../lib/utils.js';
-import { default as baseRequest } from 'request';
-import _ from 'lodash';
-import { default as moment } from 'moment-timezone';
-import cheerio from 'cheerio';
-const request = baseRequest.defaults({timeout: REQUEST_TIMEOUT});
+import log from '../lib/logger.js';
+import client from '../lib/requests.js';
+import {
+  unifyMeasurementUnits,
+} from '../lib/utils.js';
+
+import { DateTime } from 'luxon';
+
+export const parameters = {
+  CO: { name: 'co', unit: 'ppm' },
+  HUMID: { name: 'relativehumidity', unit: '%' },
+  NO: { name: 'no', unit: 'pphm' },
+  NO2: { name: 'no2', unit: 'pphm' },
+  OZONE: { name: 'o3', unit: 'pphm' },
+  PM10: { name: 'pm10', unit: 'µg/m³' },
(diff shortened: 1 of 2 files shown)


caparker:
I got it working and did some cleanup but there is still work to do on the parameters + units. 

A few notes that I wanted to share 
You were catching errors in both the `fetchStations` and the `fetchMeasurements` methods and then not rethrowing them. Each of those methods must run successfully in order for the how adapter to work. Given that and based how you designed this I would have expected either
1. No error catching in either of those methods. The try/catch in the fetchData method would catch it and we would just have to hope that the message returned was indicative of what method it happened in.
2. Catch the error in each method but then rethrow it with more detail about where it happened.

Also the repetitive use of `find` can be fixed by objectifying the array and keying it on Site_Id. 

Make sure that all the units are units we allow and if they are not we will need to fix them.

majesticio:
I corrected the use of the timezone (`'Australia/Sydney'`). The source data date times are all in local time. I made note of it in our documents for future reference

caparker (changes requested):
I am still seeing the parameter issues. Here is what I get
```
debug: Source Australia - New South Wales is "finished"
debug: Fetch results for Australia - New South Wales
debug: Started Australia - New South Wales - 1713275870399
info: 273 new measurements found for "Australia - New South Wales" in 1.932s
info: Australia - New South Wales - 58 occurrences of instance.parameter is not one of enum values: pm25,pm10,no2,so2,o3,co,bc,no,nox,pm1
info: Australia - New South Wales - 58 occurrences of instance.unit is not one of enum values: µg/m³,ppm,ppb
info: Australia - New South Wales - 27 occurrences of instance.value is not of a type(s) number
info: Finished with 1 successes and 0 failures in 2.018 seconds
debug: 93 locations of 7 parameters from 2024-04-16T13:00:00.000Z - 2024-04-16T13:00:00.000Z | 
Parameters for Australia - New South Wales {
"co":{"count":19,"errors":0,"max":0.317628,"min":-0.152889,"nulls":0},
"no":{"count":43,"errors":0,"max":0.07599178000000001,"min":-0.00096442,"nulls":0},
"no2":{"count":43,"errors":0,"max":0.0318994,"min":0,"nulls":0},
"o3":{"count":42,"errors":0,"max":0.03167671,"min":-0.001845,"nulls":0},
"pm10":{"count":55,"errors":0,"max":120.9,"min":-2.636,"nulls":0},
"pm25":{"count":35,"errors":0,"max":33,"min":0,"nulls":0},
"so2":{"count":36,"errors":0,"max":0.00209404,"min":-0.00064265,"nulls":0}}
```
Looks like we are only allowing 7 of the 8 parameters through the parameter check. Look to see if we need to remove one from the adapters parameter list or add it to our accepted parameters. 
Also, we are getting null value errors. To be consistent with other sources I think we should stop throwing those out and instead turn them into 9999s.

PR #1115, opened 2026-07-12
Fix South Africa (SAAQIS) adapter for the upgraded Envista platform

jaqkar:
## Problem

The South Africa source has been inactive since March 2024 (switched off in #1094). The cause: SAAQIS upgraded their Envitech Envista platform, and the public `/ajax/getAllStationsWithoutFiltering` endpoint the adapter scraped no longer exists (it now 302s to an error page).

The network itself is alive: 228 stations across all nine provinces (DFFE, SAWS, metros and provincial owners), with PM2.5 currently reporting at ~75 stations.

## Fix

The upgraded portal exposes an anonymous two-step flow, which this rewrite uses:

1. `POST {url}Account/GetApiFromBackToken` with an empty JSON body returns `{ key, timeout }` - a short-lived (30 min) key tied to the ASP.NET session cookie issued alongside it.
2. `POST {url}api` with `Authorization: <key>` (same cookie jar, hence `tough-cookie` as in the trinidadtobago adapter) proxies Envista REST requests: `regions`, `stations/{id}/data/latest`.

No credentials or registration are required. The proxied responses have the same shapes consumed by the israel-envista adapter, so the rewrite follows its structure.

Implementation notes:
- One flat station queue (concurrency 10) across all regions - SAAQIS times out under per-region parallelism.
- HTTP 204 means a station is currently silent, which is routine on this network; logged at debug rather than error.
- Channels are filtered on the API's own `valid` flag.
- Measurements in units that cannot be converted reliably (one station reports CO in mg/Nm3) are dropped.
- `za.json`: updated URL to https, set `active: true`, added timezone. The Richards Bay entry is untouched.

## Verification

Dry run today (`--dryrun --source 'South Africa'`):

```
566 new measurements found for "South Africa" in 31.356s
Finished with 1 successes and 0 failures in 31.615 seconds
```

566 hourly measurements from 100+ stations, including pm25 at 75 stations, no validation warnings. Sample measurement:

```json
{
  "location": "Ermelo-NAQI",
  "city": "Mpumalanga",
  "coordinates": { "latitude": -26.493348, "longitude": 29.968054 },
  "averagingPeriod": { "unit": "hours", "value": 1 },
  "date": { "utc": "2026-07-12T09:02:00Z", "local": "2026-07-12T11:02:00+02:00" },
  "parameter": "pm25",
  "value": 6.862,
  "unit": "µg/m³"
}
```

## Context

We run habistats.info, an environmental comfort index that uses OpenAQ ground monitors for air quality scoring, and noticed South Africa's absence while investigating data for Pretoria. Happy to adjust anything to fit your conventions.

### DIFF
--- a/src/adapters/southafrica.js
+++ b/src/adapters/southafrica.js
@@ -1,109 +1,262 @@
 /**
  * This code is responsible for implementing all methods related to fetching
  * and returning data for the South African data sources.
+ *
+ * SAAQIS (South African Air Quality Information System) runs Envitech
+ * Envista Web. A platform upgrade removed the old public
+ * /ajax/getAllStationsWithoutFiltering endpoint this adapter used to scrape,
+ * which is why the source went stale in March 2024. The current portal
+ * exposes an anonymous two-step flow instead:
+ *
+ *   1. POST {url}Account/GetApiFromBackToken -> { key, timeout } - a
+ *      short-lived (30 min) key tied to the ASP.NET session cookie.
+ *   2. POST {url}api with `Authorization: <key>` (same cookie jar) and a
+ *      JSON body describing the inner Envista REST request, e.g.
+ *      { url: 'regions', Method: 'GET', ... }.
+ *
+ * The proxied Envista responses ('regions', 'stations/{id}/data/latest')
+ * have the same shapes consumed by the israel-envista adapter, so this
+ * adapter follows its structure. No credentials are required.
  */
 
 'use strict';
 
-import {
-  unifyMeasurementUnits,
-  removeUnwantedParameters,
-  unifyParameters,
-} from '../lib/utils.js';
 import log from '../lib/logger.js';
-
-import { DateTime } from 'luxon';
-import _ from 'lodash';
 import client from '../lib/requests.js';
+
+import { CookieJar } from 'tough-cookie';
+import { DateTime } from 'luxon';
+import { parallelLimit } from 'async';
+import {
+  convertUnits,
+  unifyMeasurementUnits,
+  acceptableParameters,
+} from '../lib/utils.js';
 
 export const name = 'southafrica';
 
+const TIMEZONE = 'Africa/Johannesburg';
+
 /**
- * Fetches the data for a given source and returns an appropriate objectlog
- *
- * @param {object} source A valid source object
- * @param {function} cb A callback of the form cb(err, data)
+ * Fetches data from the SAAQIS Envista web API.
+ * @param {Object} source - The source configuration object.
+ * @param {Function} cb - The callback function to handle the fetched data.
  */
-
-export async function fetchData (source, cb) {
+export async function fetchData(source, cb) {
   try {
-		// source has a redirect that goes to an html page which causes issues
-    const body = await client({ url: source.url });
+    const session = await createSession(source);
+    const regionList = await envistaGet(source, session, 'regions');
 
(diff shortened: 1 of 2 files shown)


jsfenfen:
Hey @jaqkar, thanks for sending! I'm looking into how we can operationalize this. 

One question: You note the presence of mg/Nm3 units--do you recall which station uses these? I pulled the data and didn't immediately see it, presumably it wasn't reporting or was somehow left out?

jaqkar:
Hi @jsfenfen, thanks for looking into it!

The station is **ES_Motherwell** (station id 338, owner "Coega Development Corporation", in the Coega IDZ near Gqeberha). Its CO channel (channelId 5) reports in mg/Nm3; all its other channels are normal units. The station is live - it reported this morning (CO 0.51 mg/Nm3, valid=true).

The reason you didn't see it in the pulled data: the adapter drops it by design. The final filter in `getMeasurements` keeps only `['µg/m³', 'ppm', 'ppb']` after unit normalization, so the mg/Nm3 CO measurements never reach the adapter output. It only shows up in the raw Envista responses / station metadata.

One related heads-up while you're operationalizing: the 13 "ES_"-prefixed stations (including this one) appear to be low-cost multi-parameter sensor pods rather than reference monitors. For ES_Irene this was confirmed by SAWS (relayed via Prof. Rebecca Garland, who was herself surprised), and the channel suites on the others (PM1, CO2, TVOC, CH4, NH3, TSP) fit the same pattern. The adapter currently pipes them through like any other station, so depending on how OpenAQ wants to handle sensor grade you may want to tag or exclude them. Happy to help with that if useful.

jsfenfen:
Awesome, thanks for the info about the non-reference monitor scripts, passed on. 

W/r/t this PR, here's two things I noticed, would you be able to make these changes? 

1. I think CO appearing as mg/Nm3 units is changed to ppm if you pass a unitConversion argument, e.g. `stations/338/data/latest?unitConversion=true`. Just adding that arg would save the trouble of handling Nm^3...

2. I think you could grab all the station data with one call using regions? I was able to get something like this to work without timing out?  Functionally it would look something like `regions/data/latest?unitConversion=true&regionsIds=1,2,3,4,5,6,7,8,9&timebase=60` . Would you be able to do that? I think you need to keep the region / station data from the first call, but you'd be cutting out ~200+ api hits this way.

jaqkar:
Both done in 4605615:

1. **`unitConversion=true`** added. One thing worth knowing: it converts *all* gas channels to ppm/ppb, not just the mg/Nm3 one (e.g. NO2 comes back in ppb instead of µg/m3). Verified against station 338: the CO channel now arrives in ppm. A handful of stations still report CO in plain mg/m3 even with the flag; `unifyMeasurementUnits` already handles those, and I kept the final unit filter as a guard against anything residual (it now only drops genuinely exotic units like the ng/m3 mercury channels, which aren't accepted parameters anyway).

2. **Single bulk call** via `regions/data/latest?unitConversion=true&regionsIds=...&timebase=60`, with region ids taken from the initial `regions` call (which is still needed for station metadata: names, coordinates, region). This drops ~230 per-station requests down to 2 requests total. Stations with no current data are simply absent from the bulk response, so the per-station 204 handling went away too.

Dry run after the change: 587 measurements across 119 locations in ~5.5s, 0 failures (was 566 measurements via the per-station loop). CO now included at 40 locations.

russbiggs:
Thanks @jaqkar for your contribution!

